PR #3, opened 2018-01-18
ST-Link support

johnjiabinzhang:
This pull request adds support for ST-Link tools.
It:
 - adds a debugger type stlink-gdb
 - provides STLink implementation based on JLink
 - adds device name to SVD file mapping for STM32L4x6 MCUs

Current limitations:
 - SWO is not supported
 - launch only

It also fixes some problem with npm script and launch config, see 76fb672.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -13,8 +13,8 @@
 			"sourceMaps": true,
 			"outFiles": [
 				"${workspaceRoot}/out/**/*.js"
-			]
-			// "preLaunchTask": "npm"
+			],
+			"preLaunchTask": "npm"
 		},
 		{
 			"name": "JLink debug server",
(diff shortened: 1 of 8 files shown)


Marus:
Regarding the "CPSR" register.

It is not actually technically correct that it is another name for XPSR. CPSR is actually a register from the older ARM7TDMI architecture (may be in other ARM architectures as well) that is functionally similar, although not identical, to the XPSR register from ARMv6-M and ARMv7-M architectures.  I have left the work around you added in there as the st-util GDB server incorrectly identifies that register as CPSR.

That GDB server seems to have some issues with registers in general, in addition to miss-identifying the XPSR as CPSR, it doesn't provide the PSP and MSP banked SP registers, nor the PRIMASK, FAULTMASK, BASEPRI or CONTROL registers. And it incorrectly says that Cortex-M3 have registers F0-F7 and FPS - which do not exist.

PR #11, opened 2018-01-30
Polish memory view

johnjiabinzhang:
This pull request takes portions of code from the [Hexdump](https://github.com/stef-levesque/vscode-hexdump/) extension and integrate them into cortex-debug. The result is a finer looking memory view with configurable settings (additional `cortex-debug.memoryDump` entry).

This also fixes a bug in the current `v0.1.11` release where `vscode.Uri.parse()` is required to parse the examine memory string into URI (possibly due to a recent VSCode API change).

Currently the hovering display feature of Hexdump is not included. Currently it's not a high priority. I may do a future PR for this.

Copyright notice for the potion of code taken from Hexdump is also included.

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
     ],
     "categories": [
         "Debuggers"
-    ],
+    ], 
     "contributes": {
         "configuration": {
             "type": "object",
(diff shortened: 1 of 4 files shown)


Marus:
Ok - Had a look at this - and there are a few things that I don't like with how this is implemented/works currently:

* Shouldn't use the **.hexdump** extension. I was previously using it as I was making use of the hexdump extension for the syntax highlighting (I was intending to change this for the next release actually and define my own syntax highlighting as it couldn't make use of the hover inspector and most other functionality). In this case you are defining a new language for .hexdump - which makes the usability of both extensions worse as you have to manually switch between which language it is using for the .hexdump extension to get the appropriate functionality depending on how you are using it at the current time.
* Without the hover inspector the **littleEndian** setting serves no purpose - that was used to determine how to decode the bytes for the inspector. I believe the **charEncoding** setting is also unused.
* I don't like the presentation if you are starting at an address that is not aligned with the set width value. Previously it would start the first line at the nearest preceding address that is aligned to the width, leaving gaps for the memory that was not read. The big advantage of that presentation is that it makes the mental math for determining the address of a particular byte much easier.

johnjiabinzhang:
Thank you for the feedback Marus. I see now that I didn't test it properly. I had Hexdump extension on, and there were some interference from there. Now when I turned it off, it makes more sense what it's doing.

I see your points now. I think I will close this pull request and rework on the memory view. I will base it on your code this time instead of using `hexy`.
I will open another pull request soon.

Marus:
I think the concepts of what you were trying to do were good, and would be welcomed - but I think the actual implementation needs a bit more refinement before it is ready to go.

johnjiabinzhang:
Sure thing. I'm working on it now. Will let you know soon. :)

PR #16, opened 2018-02-06
Bug Fix: non-resolved promise

johnjiabinzhang:
the promise in `GDBServer.init()` is not resolved in the main branch of `if`.
This causes a time out problem and the server sub-process is terminated.
This bug became apparent in Node v8.9.4, yet works with older version of Node (e.g. `v8.9.1` on my macOS).

### DIFF
--- a/src/backend/server.ts
+++ b/src/backend/server.ts
@@ -26,9 +26,8 @@ export class GDBServer extends EventEmitter {
 				this.process.on('exit', this.onExit.bind(this));
 				this.process.on('error', this.onError.bind(this));
 			}
-			else { // For servers like BMP that are always running directly on the probe
-				resolve();
-			}
+			resolve();
+			// For servers like BMP that are always running directly on the probe
 		});
 	}
 


Marus:
This fix is not correct; if you are getting a timeout that would mean it isn’t detecting the GDB server as having launched correctly (the resolve and reject callbacks are stored in the initResolve and initReject instance variables and called later when it has detected). I’ll have a look to see why it may not be detecting the server as having initialized properly.

Marus:
Can you send me your launch.json configuration and a log of the "Adapter Output" output window for a case where you are receiving the timeout? That should hopefully help me identify why it is not detecting the GDB server as having launched properly - and why it doesn't actually perform the resolve action.

johnjiabinzhang:
Hi Marus, you are right. This is not the right fix. It seems that the problem is the redirected `stdout` and `stderr` handler function is not invoked on my Windows based testing environment.
Everything works fine on macOS however. I will drill this down more and see what the problem is.

Marus:
Ok - If I have a chance I'll look into it - could be something strange with the handling of the output on Windows that is causing it to not receive the input. Will let you know if I find anything.

johnjiabinzhang:
I'm banging my head on this thing at the moment. There is something strange about handling the output on Windows. The `onStdOut` callback is never called.

This should be a very significant problem, I'm surprised that it still exists even though it seemed to have been fixed back in 2015. I wonder if wrapping the spawning in a resolution-delayed promise makes any difference.

I will look into this more tomorrow.

PR #22, opened 2018-02-17
make memory view live

johnjiabinzhang:
track memory view docs, when a stop event occurs, update their contents.

### DIFF
--- a/src/frontend/extension.ts
+++ b/src/frontend/extension.ts
@@ -34,9 +34,11 @@ class CortexDebugExtension {
 
 	private peripheralProvider: PeripheralTreeProvider;
 	private registerProvider: RegisterTreeProvider;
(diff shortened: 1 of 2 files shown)


Marus:
I think this all looks good, I’ll give it a final test and include it in the next release.

johnjiabinzhang:
Great!

johnjiabinzhang:
Also, I'm noticing that in `memory_content_provider.ts` line `25` and `46`, it is subtracting `0x08000000` from the line address. Is there any particular reason why? I don' think it should be there.

Marus:
No, it shouldn’t be there. At some point I was experimenting with trying to integrate with the hexdump extension - must have accidentally committed something. I’ll fix that.

PR #29, opened 2018-03-14
Added -s argument to openocd configuration

KaDw:
It's handy to have -s argument which specifies dir to search for config files and scripts.
Then while writing a .cfg script you can have a specific configuration in your project and just include needed scripts with following command ```source [find interface/very_useful.cfg] ``` 

eg.
```
source [find interface/stlink.cfg] 

set WORKAREASIZE 0x8000

transport select "hla_swd"

# Enable debug when in low power modes
set ENABLE_LOW_POWER 1

# Stop Watchdog counters when halt
set STOP_WATCHDOG 1

# STlink Debug clock frequency
set CLOCK_FREQ 4000

reset_config srst_nogate
set CONNECT_UNDER_RESET 1

source [find target/stm32f4x.cfg]
```

There are also some other ways to add search dir to OpenOCD:
```
Configuration files and scripts are searched for in
1. the current directory,
2. any search dir specified on the command line using the -s option,
3. any search dir specified using the add_script_search_dir command,
4. $HOME/.openocd (not on Windows),
5. a directory in the OPENOCD_SCRIPTS environment variable (if set),
6. the site wide script library $pkgdatadir/site and
7. the OpenOCD-supplied script library $pkgdatadir/scripts
```


Marus:
Overall this looks good - but I can think of one issue you'll need to fix. There should be something added to src/frontend/configprovider.ts so that if they haven't specified that configuration parameter it will work - as otherwise when you try to do the forEach on that config parameter in openocd.ts it will fail. (you can see a bunch of similar ones around line 67).

Once that is done I'll include it in the next release - thanks for the contribution.

Marus:
Ok - that looks good now. I’ll give a final test and merge it in for the next update. Thanks.

PR #50, opened 2018-04-30
Fix issue with enumeratedValue without "description" and peripheral registers without "register"

ivankravets:
Subj


ivankravets:
Sure! See attached SVD.

[ATSAMD21G18A.svd.zip](https://github.com/Marus/cortex-debug/files/1963342/ATSAMD21G18A.svd.zip)

PR #120, opened 2019-01-21
Option to show a particular SWO output window during debugger startup

ssimek:
Hi Marcel,

I'd love if you could include the option to show a SWO output window during startup

Thanks,

Stefan


Marus:
Hey @ssimek - Will include that in the next release.

PR #124, opened 2019-02-04
Alternate memory view

andenore:
Added an alternate memory view using webview.

This view automatically reads more memory as you scroll, and also let's you select different memory layouts (ascii, uint8, uint16, uint32).

I haven't seen it under the current issues, but I've been missing a simpler webview myself.

### DIFF
--- a/memoryview/index.html
+++ b/memoryview/index.html
@@ -42,6 +42,7 @@
 	let g_vscode = undefined;
 	let g_memoryData = [];
 	let g_readInProgress = false;
+	let g_scrollY = 0;
 	let DataFormatEnum = {
 		"ascii" : 0,
 		"uint8" : 1,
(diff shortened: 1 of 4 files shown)


Marus:
I'll take a look at the PR this coming weekend and see about merging it into the next release if everything seems to be working well.

danebou:
While we're add it, could we potentially expand this view for memory editing?

andenore:
I was thinking about adding memory editing, I don't think it should be too difficult. I can have a go at it, but maybe that could be another pull-request if this gets accepted?

Marus:
Sorry - been busy with other things - myself and/or @haneefdm who has been helping me with the project recently will try to take a look at this soon.

haneefdm:
Yes, I will take a look. On Thu/Fri

Marus:
At the moment I think both myself and @haneefdm felt this wasn't a great fit for the project. While it did make some functionality additions we felt this came with some fairly significant complexity by using a web view. I think we may look at eventually adding some of the features provided in the web view in another manner through the current memory view system - but for now at least we aren't going to merge this in.

haneefdm:
@andenore I learned a bit about webview thanks to you. The auto-scroll/fill was nice too. But the window reset itself to the top every time it changed focus/exposed and re-filled all the data losing the original benefit. I used to call the auto-scrolling/fill as lazy-filling. We can try to do the same in the existing memory view and also supports the additional formats.

PR #127, opened 2019-02-15
Add optional GDB port scan start and end range

slademcthornbody:
Add optional GDB port scan start and end range

### DIFF
--- a/package.json
+++ b/package.json
@@ -559,15 +559,26 @@
                                 "required": [],
                                 "type": "object"
                             },
-                            "portRangeStart": {
-                                "default": null,
-                                "description": "Start range for GDB port scan. Optional.",
-                                "type": "string"
+                            "fixedPortRangeStart": {
+                                "type": "integer",
+                                "default": -1,
+                                "minimum": -1,
+                                "maximum": 65535,
+                                "description": "Fixed port for GDB port"
                             },
-                            "portRangeEnd": {
-                                "default": null,
-                                "description": "End range for GDB port scan. Optional.",
-                                "type": "string"
+                            "portScanRangeStart": {
+                                "type": "integer",
+                                "default": 50000,
+                                "minimum": 1024,
+                                "maximum": 65535,
+                                "description": "Start range for GDB port scan"
+                            },
+                            "portScanRangeLength": {
+                                "type": "integer",
+                                "default": 2000,
+                                "minimum": 0,
+                                "maximum": 65535,
+                                "description": "Start range for GDB port scan"
                             },
                             "ipAddress": {
                                 "default": null,
@@ -1590,5 +1601,5 @@
         "vscode:prepublish": "tsc -p ./",
         "watch": "tsc -watch -p ./"
     },
-    "version": "0.2.4"
+    "version": "0.2.6"
 }
--- a/package.json
+++ b/package.json
@@ -225,16 +225,6 @@
                                 "description": "Target Device Identifier",
                                 "type": "string"
                             },
-                            "portRangeStart": {
-                                "default": null,
-                                "description": "Start range for GDB port scan. Optional.",
-                                "type": "string"
-                            },
-                            "portRangeEnd": {
-                                "default": null,
-                                "description": "End range for GDB port scan. Optional.",
-                                "type": "string"
-                            },
                             "rtos": {
                                 "default": null,
                                 "description": "RTOS being used. For JLink this can be FreeRTOS, embOS or the path to a custom JLink RTOS Plugin library. For OpenOCD this can be eCos, ThreadX, FreeRTOS, ChibiOS, embKernel, mqx, or uCOS-III.",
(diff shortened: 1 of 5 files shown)


noppej:
@slademcthornbody Do you think this will address my requirement from [https://github.com/Marus/cortex-debug/issues/117#issuecomment-470650518](url)

slademcthornbody:
I don't know, youd need to try it. My change still results in a call to portastic, but based on my changes it should be trivial to remove this completely and define a fixed port for operation

package.json
```diff
@@ -1080,6 +1090,16 @@
                                 "pattern": "^(25[0-5]|2[0-4][0-9]|1[0-9][0-9]|[1-9]?[0-9])\\.(25[0-5]|2[0-4][0-9]|1[0-9][0-9]|[1-9]?[0-9])\\.(25[0-5]|2[0-4][0-9]|1[0-9][0-9]|[1-9]?[0-9])\\.(25[0-5]|2[0-4][0-9]|1[0-9][0-9]|[1-9]?[0-9])$",
                                 "type": "string"
                             },
+                            "portRangeStart": {
+                                "default": null,
+                                "description": "Start range for GDB port scan. Optional.",
+                                "type": "string"
```
go2sh: You can use `integer` here and get proper validation for free and you don't have to convert it to int.

src/common.ts
```diff
@@ -88,6 +88,8 @@ export interface ConfigurationArguments extends DebugProtocol.LaunchRequestArgum
     executable: string;
     servertype: string;
     serverpath: string;
+    portRangeStart: string;
```
go2sh: Change to number here.

src/gdb.ts
```diff
@@ -199,7 +199,18 @@ export class GDBDebugSession extends DebugSession {
         this.debugReady = false;
         this.stopped = false;
         
-        portastic.find({ min: 50000, max: 52000, retrieve: this.serverController.portsNeeded.length }, '0.0.0.0').then((ports) => {
+        var IntPortRangeStart = 50000;
+        if (this.args.portRangeStart != null)
```
go2sh: You can remove the check then and use the arg value directly.

package.json
```diff
@@ -1080,6 +1090,16 @@
...
+                            },
+                            "portRangeEnd": {
+                                "default": null,
```
go2sh: Set it to `-1`.

package.json
```diff
@@ -1080,6 +1090,16 @@
                                 "pattern": "^(25[0-5]|2[0-4][0-9]|1[0-9][0-9]|[1-9]?[0-9])\\.(25[0-5]|2[0-4][0-9]|1[0-9][0-9]|[1-9]?[0-9])\\.(25[0-5]|2[0-4][0-9]|1[0-9][0-9]|[1-9]?[0-9])\\.(25[0-5]|2[0-4][0-9]|1[0-9][0-9]|[1-9]?[0-9])$",
                                 "type": "string"
                             },
+                            "portRangeStart": {
+                                "default": null,
```
go2sh: Set it to a proper default.

src/gdb.ts
```diff
@@ -199,7 +199,18 @@ export class GDBDebugSession extends DebugSession {
...
+        {
+            IntPortRangeStart = parseInt(this.args.portRangeStart, 10);
+        }
+        var IntPortRangeEnd = IntPortRangeStart + 2000;
+        if (this.args.portRangeEnd != null)
```
go2sh: Check for a value bigger the portRangeStart, else use your 2000 default

package.json
haneefdm: Adding to what @go2sh said, you can also specify a proper range. A better spec. is to ask for a number greater than 0 (0 being special) up to the max and you get auto-validation for free. Maybe we back off on the max a bit., perhaps by 2000 to match what we have.
                           
```
"portRangeStart": {
    "type": "integer",
    "default": 50000,
    "minimum": 1,
    "maximum": 65535
}
```

package.json
haneefdm: I don't think a portRangeEnd is needed at all. Extension scans for 2000 ports starting from "portRangeStart" and we are done. If if you want to have a "portRangeEnd", you may want to specify as "numPortsToScan", the default being 2000 like it is now. and it better be a positive integer. Some targets need a minimum set of targets, so having this property creates additional error checking to make sure we satisfy the max. 2000 is a huge set.

slademcthornbody:
Thanks for the feedback - agree that changing the scan parameters is a bit messy when Portastic isn't working properly so I have added a fixed port param to override the port scan behavior completely. Have tried to address feedback above also - first time playing with njs. See new commit. Cheers

package.json
haneefdm: I think `fixedPortRangeStart` default value of -1 is being used to see if a user wants a fixed one or a scanned one. And, it takes precedence over `portScanRangeStart`. May I recommend using type to be
`"type": ["integer", "null"]`
or using the `anyOf` and default can be null. And, I think you can still have reasonable min/max in case of an integer.

The so-called spec I found here https://cswr.github.io/JsonSchema/spec/multiple_types/

slademcthornbody: Have changed this also

Marus:
I'm going to close this PR - I'm working on an update to the port selection to (hopefully) make detection of unused ports more reliable, and will also support the specification of a starting port to use.

PR #136, opened 2019-03-11
Pyocd updates

pelrun:
Allow use of cmsis-pack files for additional target support.

Update parsing of gdbserver output for latest versions of pyOCD.


Marus:
Just wanted to check with you on this to see what the minimum required version of PyOCD would be to support these new options so that it can be included in the documentation. Also can you update to resolve the conflicts (should just be around the initMatch value which I had also changed in V0.2.5)

pelrun:
The custom CMSIS-Pack support came in at PyOCD v0.16.0.

Marus:
Thanks for the info @pelrun - will try to have a final look at this tomorrow and should likely go out with the next release. Thanks for the contribution.

PR #138, opened 2019-03-12
Update STM32L0x SVD to v1.2

clementperon:
Archive found here: https://www.st.com/content/st_com/en/support/resources/resource-selector.html?querycriteria=productId=SS1817$$resourceCategory=cad_models_and_symbols$$resourceType=svd


Marus:
This will go out in a release this weekend.

PR #146, opened 2019-04-04
Adding ability to set port scanning range 

blckmn:
in case there is a permissions issue / blocked / inuse by other app


blckmn:
Already in the pipeline. Duplicate of https://github.com/Marus/cortex-debug/pull/127

Issue #156, opened 2019-04-13
Random crashes in the debug adapter


haneefdm:
It happens when I stop a debug session once in a blue moon (launch or attach does not matter). Very difficult to reproduce, but it kills the debug adapter. I am using the adapter in server mode (debugging stuff). This is the stack trace you see...

```
Error: kill ESRCH
    at process.kill (internal/process/per_thread.js:198:13)
    at Timeout.stop.setTimeout [as _onTimeout] (/Users/hdm/cypress/vscode/cortex-debug/src/backend/mi2/mi2.ts:224:47)
    at ontimeout (timers.js:436:11)
    at tryOnTimeout (timers.js:300:5)
    at listOnTimeout (timers.js:263:5)
    at Timer.processTimers (timers.js:223:10)
```
I have not investigated it much but it may be a condition where the server/gdb already quit and we are trying to kill it. Just wanted to capture it as an issue.

How can I help debug it? Or see how this can even happen. If this is a known issue please let me know. I can help in whatever way you want.

minkir014:
This happens when you launch or when you stop the debugger.

haneefdm:
When I hit the stop debugging button (ending a debug session). The button next to restart. From what In the source code, we ask gdb to exit and then for good measure, a second later also try to send a kill signal if not already quit. There may be a race condition between the three lines of code at the end of stop/detach. In src/backend/mi2.ts. A try/catch around process.kill() may be necessary
```
    public stop() {
        const proc = this.process;
        const to = setTimeout(() => { process.kill(-proc.pid); }, 1000);
        this.process.on('exit', (code) => { clearTimeout(to); });
        this.sendRaw('-gdb-exit');
    }

    public detach() {
        const proc = this.process;
        const to = setTimeout(() => { process.kill(-proc.pid); }, 1000);
        this.process.on('exit', (code) => { clearTimeout(to); });
        this.sendRaw('-target-detach');
    }
```
When the gdb adapter dies, future debugging sessions will stop working.

minkir014:
You can start a pull request as you started before that and I would support you. Make try/catch expressions as you said and try it and tell me the result.

haneefdm:
@minkir014, I already have a PR #151  in the queue from more than 3 weeks ago. I am trying not to do additional PRs until that one is accepted. I am trying to document issues I find to go back and fix one at a time. I know the fixes for almost everything I find but also helps to get feedback.

[...]

PR #169, opened 2019-05-25
More reliable auto tcp port selection, tslint cleanup on some files, mostly mine

haneefdm:
There are a few new files. The core is not affected much though and easily revertable

**Summary:**
- My new method of finding free TCP ports is much more reliable on all platforms (Mac/Linux/Windows)
- Lot less dependence on what gdb server output looks like to know if it started successfully. This method is probably a guaranteed method.
- Totally compatible with existing ways of knowing when the gdb server started and ready.
- Some housekeeping changes. Enforce indentation/tab usage for the project. It makes pull requests easier and keeps `tslint` happy. Only on files I touched though.
- Unit tests to hopefully find when OS updates break Cortex-Debug port-scanning
- Perhaps fixes issue #156. I can't tell because it is so random. But, it is a safe fix.

**Testing:**
- I tested a lot but need to get some independent testing to see if your company firewall rules cause issues.
- I tested with OpenOCD, but the way the new code works, it should not affect any other servers. In fact, should be safer.
- There is one corner case, where when we are testing to see if a server opened a port, a race condition could happen between the test and the gdb-server starting properly. One in a million kind of thing, but it can happen. Still more reliable and faster than what we have. I know a better way but not through `Node.js`

**Contributors to Cortex-Debug**
A new npm module `tcp-port-used` is needed to build, so please do an `npm install` which might bring you to a whole bunch of `tslint` issues when you open a file in VSCode. Everything still works though, no fatal errors.


haneefdm:
One issue I ran into was that what `tslint` wants and the VSCode editor behaves are not in sync in many different ways (spaces, tabs, quotes, indentation, etc.)

PR #172, opened 2019-05-31
Patch openocd timeout until proper solution is found

adammunich:
We should properly check the status of the openocd port but for now let's patch this so we don't waste people's time.


Marus:
This was an intentional change. We recommend updating to a newer version of OpenOCD (either building from source yourself or using the releases from GNU MCU Eclipse (https://gnu-mcu-eclipse.github.io/openocd/install/).

Additionally with the next release you'll also be able to override the default detection RegEx if you are forced to use the very old version of OpenOCD that doesn't have the default output.

Issue #155, opened 2019-04-12
Feature Request: Formalize toolchain and server paths customizable by platform


haneefdm:
I believe the code is already there to do everything needed. But it is not declared in package.json schema and I believe there is also an error for `armToolchainPath` in there as well.
```
            "osx": {
                "toolchainPath": "/Applications/ModusToolbox_2.0/tools/gcc-7.2.1-1.0/bin",
                "serverpath": "/Applications/ModusToolbox_2.0/tools/openocd-2.2/bin/openocd",
                "armToolchainPath": "xxxyaaz"  // Has no effect
            },
```
The `toolchainPath` and `serverpath` actually do work but not specified in package.json and Intellisense. **Could we formalize these features please**?

We have team members working on multiple OSes and switching between them, being able to switch easily can be really helpful. settings.json (local or users home dir) does not have OS specific overrides (I think).

It is unfortunate that `serverpath` is not camelCase and neither is `toolchainPath` strictly speaking, so we have to stick with them for compatibility.

The `armToolchainPath` property shows up in IntelliSense but has no effect in launch.json; it is really for settings.json? I see in changelog v0.1.20 that a change was made to support per-launch basis. Not seeing evidence of that.

# Fixes?
I can fix this by editing the package.json. Don't want to rename config. names though unless there is a consensus. I volunteer to help with the approval of the author/team. Please comment.

minkir014:
You can use Microsoft cpp extension for intellisense. I can help you with this. And use this extension for debugging.

haneefdm:
@minkir014, I already use the Microsoft cpp extension and also providing bug fixes to those repositories (MIEngine and vscode-cpptools) as well. This is not about editing C/C++ code. I am talking about when editing launch.json for my embedded c++ project for the `cortex-debug` debugger configuration.

I debug using both extensions (vscode-cpptools and cortex-debug)

minkir014:
If you have the solution to fix that. Start a PR.

Marus:
@haneefdm - any option in launch.json can be put in an OS specific block already - and you should be getting auto-completion. The options that are in your user/workspace settings don't have that - as those types of settings are generally specific to the machine and not the project. 

[...]

PR #175, opened 2019-06-24
Allow toolchain prefix to be configured per launch configuration

ryanplusplus:
This is currently exposed in the user/workspace configuration, but we have projects that use multiple toolchains in the same workspace because there are multiple build targets. This would help us by allowing us to select a different toolchain variant per launch configuration.

Before you ask, the reason that we even need the prefix is that we're using `cortext-debug` for both ARM Cortex microcontrollers and non-ARM microcontrollers. If we ignore the intellisense warnings in the configuration and smooth over the prefix problems this works as long as we name our alternative toolchains "correctly". I know this is a weird use case but `cortex-debug` is much better than any of the alternatives and being able to use it with our strange project configurations without hacks would be much appreciated.

### DIFF
--- a/package.json
+++ b/package.json
@@ -759,6 +759,11 @@
                                 "description": "This setting can be used to override the armToolchainPath user setting for a particular launch configuration. This should be the path where arm-none-eabi-gdb and arm-none-eabi-objdump are located.",
                                 "type": "string"
                             },
+                            "toolchainPrefix": {
+                                "default": null,
+                                "description": "This setting can be used to override the toolchainPrefix user setting for a particular launch configuration.",
+                                "type": "string"
+                            },
                             "executable": {
                                 "description": "Path of executable",
                                 "type": "string"
--- a/package.json
+++ b/package.json
@@ -253,6 +253,11 @@
                                 "description": "This setting can be used to override the armToolchainPath user setting for a particular launch configuration. This should be the path where arm-none-eabi-gdb and arm-none-eabi-objdump are located.",
                                 "type": "string"
                             },
+                            "toolchainPrefix": {
+                                "default": null,
+                                "description": "This setting can be used to override the toolchainPrefix user setting for a particular launch configuration.",
+                                "type": "string"
+                            },
                             "executable": {
                                 "description": "Path of executable",
                                 "type": "string"
--- a/src/frontend/configprovider.ts
+++ b/src/frontend/configprovider.ts
@@ -89,7 +89,9 @@ export class CortexDebugConfigurationProvider implements vscode.DebugConfigurati
         if (!config.toolchainPath) {
             config.toolchainPath = configuration.armToolchainPath;
         }
-        config.toolchainPrefix = configuration.armToolchainPrefix || 'arm-none-eabi';
+        if (!config.toolchainPrefix) {
+            config.toolchainPrefix = configuration.armToolchainPrefix || 'arm-none-eabi';
+        }
         
         config.extensionPath = this.context.extensionPath;
         if (os.platform() === 'win32') {


Marus:
@ryanplusplus - Out of curiosity - what other microcontrollers are you currently using it with? While obviously the peripheral registers system isn't likely to be of much use currently (although if they are memory mapped peripherals and you were ambitious enough you could theoretically write an SVD definition for them) do the other features like core registers, memory views, etc all work well?

ryanplusplus:
> The package.json changes would need to be replicated in both the attach and launch sections (this is a side effect of their format which does not allow us to declare all those properties in a shared section unfortunately) - otherwise the editing intellisense will only work with one. Other than getting that fixed I think this looks fine and can go out with the next release likely.

This is my first time working with a VS Code extension and I'm mostly guessing my way through. I very much appreciate the feedback and will work on those changes tomorrow.

> @ryanplusplus - Out of curiosity - what other microcontrollers are you currently using it with? While obviously the peripheral registers system isn't likely to be of much use currently (although if they are memory mapped peripherals and you were ambitious enough you could theoretically write an SVD definition for them) do the other features like core registers, memory views, etc all work well?

Right now the primary alternative target is Renesas RX13x and RX2xx microcontrollers. We haven't been using `cortex-debug` for very long, but the core registers, breakpoints, locals, etc. appear to be working correctly with the exception of static variables (which, to be fair, I haven't seen working on Cortex M targets either). Provided we don't run into any show stoppers I'm planning to attempt to make SVDs for the parts we use because they do use memory-mapped peripherals and I could likely write a tool to machine-generate it from the register header files. It would be great if long term this project supported these alternative targets, but for now I'm happy enough with the support we have given that we can make it work by ignoring intellisense warnings.

Incidentally, supporting alternative microcontrollers also means that #155 would be very useful for us. As a general rule we pin JLink versions (along with toolchain versions) per project and automatically pull the right version during build. This is helpful because we've run into issues where RX support was broken on a JLink release that fixed issues for other parts. Being able to manage the JLink server on a project-by-project basis instead of workspace or machine-wide allows us to sidestep these issues (in addition to removing external dependencies from our projects). If you're amenable to the changes proposed in #155 I'd be happy to expose a first class configuration item for the server path, but otherwise we're able to use the undocumented/accidental `serverpath`.

haneefdm:
@ryanplusplus About #155. I found an issue  (VSCode weirdness) as to why certain items defined in launch.json to be OS specific overrides. I will be doing a PR tomorrow -- I have to test on all three platforms. I will take care of the 'serverpath' as well, with @Marus permission.

haneefdm:
In my lifetime, had many reasons to fix a project to one particular version of gcc/gdb/libc. Just two weeks ago, I found a crash in gdb and only 8.2.x will work for what I am doing. Was working fine in 7.x, regressed in 8.1.x and then finally fixed in 8.2+ (which is released but not all distros have picked it up).

If you are curious, doing a `-data-list-registers` or `info registers` crashed gdb for x64. Had issues with g++ parser as well in the past.

ryanplusplus:
> I will be doing a PR tomorrow -- I have to test on all three platforms. I will take care of the 'serverpath' as well, with @Marus permission.

Thanks @haneefdm, much appreciated.

ryanplusplus:
@marus this now works for both launch and attach configurations.

Marus:
Thanks @ryanplusplus - we'll include this in the next release. And as @haneefdm mentioned he is working on a fix for #155 - so that will likely also come in the next release.

Marus:
Oh and @ryanplusplus - regarding the global variables - there are a few features, mainly some of the disassembly stuff and global variables, that rely on the output of the objdump tool (originally had to be `arm-none-eabi-objdump` but will now be able to have the custom prefix). 

So I can see two things that may be happening here that would cause them to not work (for both arm targets and for your Renesas ones)

1. It could be failing to execute the objdump tool so it just can't load the symbol information to find the globals - so you can check to see if that tool is running properly.
2. It could have an issue with parsing the output from objdump. If you want to post the output of `<prefix>-objdump --syms <executable>` for a small sample (with global variables) in a new issue I can see if I can see if there is anything that seems like it wouldn't be parsed correctly.

ryanplusplus:
@marus when I aliased `rx-elf-gdb` to get it working without changing the prefix I didn't do the same for `r-elf-objdump` so that's almost certainly the problem. I'll check it in the office tomorrow and let you know.

ryanplusplus:
@Marus I tried again after changing `rx-elf-objdump` to `arm-none-eabi-objdump` and still was unable to see globals or statics in the Variables view. Neither of these is a big deal for us because we can manually add watches, but I included details in https://github.com/Marus/cortex-debug/issues/179 if you want to investigate.

PR #176, opened 2019-06-25
Minor changes to port-scanner, better handle platform specific overrides

haneefdm:
Fix for Issue #155 Should also fix Issue #177

Only affects people using VPN tunnels. Sometimes VPNs open a special IP address for a tunnel and creating a server on that port fails and we discard the entire port because of one IP address alias failing.

We now only look for IPv4 and internal (loopback) IP aliases to be free. Greater chances of success and fewer chances of errors due to VPN weirdness.

I also added another function waitForPortOpenOSUtil(). This works based on what the OS tells us what the port is doing rather than create a server or a client. Totally unobtrusive but slow. I haven't pushed the change to server.ts to use this yet but it is in my code base.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ node_modules
 .vscode-test
 .DS_Store
 .sass-cache
+/dist
--- a/package.json
+++ b/package.json
@@ -253,6 +253,11 @@
                                 "description": "This setting can be used to override the armToolchainPath user setting for a particular launch configuration. This should be the path where arm-none-eabi-gdb and arm-none-eabi-objdump are located.",
                                 "type": "string"
                             },
+                            "serverpath": {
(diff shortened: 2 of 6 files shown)


haneefdm:
Modified configprovider.ts to flatten OS specific stuff before processing/adjusting the properties.

Added 'serverpath' to launch.json

Marus:
@haneefdm - Will try to have a look at this over this coming weekend to get it ready to go in the next release.

Marus:
Also - looks like there were some minor conflicts as a result of merging in @ryanplusplus's changes to expose the toolchain prefix in the launch.json file

haneefdm:
I think it is just the .gitignore.

Changes in package.json, are in separate areas I think. I only added the
'serverpath'

PR #177, opened 2019-06-27
Fix build on case sensitive file systems

ryanplusplus:
When testing #175 I ran into issues building because I'm running on EXT4 which is a case sensitive file system. This caused a few imports to fail because of a difference between the filenames on disk and the filenames referenced in the source.

With these changes I'm able to build cleanly on a case sensitive file system.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+dist
 out
 node_modules
 *.vsix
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -1,3 +1,4 @@
+dist
```
ryanplusplus: I also noticed that the build output was not gitignored

Marus:
Thanks for these little fixes @ryanplusplus. When I recreated the grapher source (original source had be lost somewhere along the way with my old computer dying with only the web-packed version surviving) I hadn't noticed that I had changed the casing - and while macOS is case preserving it is not case-sensitive (by default).

Issue #180, opened 2019-07-05
Feature Request: Separating launch/restart/attach override commands from SWO config.


haneefdm:
I was getting ready to use the new override commands for launch/restart/attach, but ran into an issue.

Currently, the built-in commands for these actions also do the SWO configuration. It could be a lot more useful if we separated the SWO configuration from the basic startup commands, or I have to kinda give up the SWO functionality (or replicate it) to use the overrides.

I can do the work for all gdb-servers, if this is acceptable. The overrides can then be a lot more useful. Opinions?

Marus:
This is a very good idea actually- didn’t think about that when I added the the override settings, but you are correct, as currently implemented they would likely break the SWO functionality. Would be happy to look at a PR for this.

haneefdm:
Thanks, @Marus. I will get to work on it tomorrow.

haneefdm:
This released in 0.3.3

PR #184, opened 2019-07-11
Separate swo config. from launch/attach/restart commands

haneefdm:
This PR will address #180 and issues related to attach/detach

- attach/detach related changes (mi2.ts && gdb.ts). If it is an attach session, we never set a flag to remember that although we test for it. For a detach to succeed, we have to interrupt first if not in a stopped state or else detach will fail. Send a normal detach request and then ask gdb to exit. Give gdb to detach/exit before killing server, cleanup, lock files, etc. The `disconnectRequest` is a bit more graceful now.

- Separate swo configuration commands from launch/attach/restart commands. Also, do the swo configuration **after postXXXcommands** since swo may not succeed until the launch is completely done.

As an aside, should we remove any breakpoints before a detach? I don't have a good test for it but I have a feeling neither gdb or the gdb-server removes the breakpoints. A safe way for users to detach is to pause, remove all breakpoints, continue and then request a detach. A better way to shutdown the server could be `monitor shutdown` for openocd, but such a command is not universal. Other gdb-servers use `monitor exit`

Note: I verified swo related changes by verifying the commands are being generated properly. I don't have a good SWO setup to fully verify.

### DIFF
--- a/src/backend/mi2/mi2.ts
+++ b/src/backend/mi2/mi2.ts
@@ -237,10 +237,8 @@ export class MI2 extends EventEmitter implements IBackend {
     }
 
     public detach() {
-        const proc = this.process;
-        const to = setTimeout(() => { this.tryKill(); }, 1000);
-        this.process.on('exit', (code) => { clearTimeout(to); });
-        this.sendRaw('-target-detach');
+        this.sendCommand('target-detach');
+        this.stop();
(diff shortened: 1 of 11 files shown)


Marus:
Will try to test this out this weekend and get it merged in finally.

PR #186, opened 2019-07-18
Add 'stopOnEntry' configuration

kaegi:
I'd like to be able to (re)start debugging without the debugger pausing automatically. This way the debugger can advance (possibly to the first user-set breakpoint) without pressing `continue`. Such behavior seems to be called `stopOnEntry` (false) in the VSCode world.

This also presents an alternative to `runToMain`, because the user can add a breakpoint at the start of the main function, which results in the same behavior.

I think `stopOnEntry=false` should be the default value, but for backwards compatibility it is set to 'true'.

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,7 +19,10 @@
                     "description": "Path to the GCC Arm Toolchain (standard prefix is \"arm-none-eabi\" - can be set through the armToolchainPrefix setting) to use. If not set the tools must be on the system path. Do not include the executable file name in this path."
                 },
                 "cortex-debug.armToolchainPrefix": {
-                    "type": ["string", "null"],
+                    "type": [
+                        "string",
+                        "null"
+                    ],
                     "default": "arm-none-eabi",
                     "description": "The prefix to use for the arm toolchain - the standard release from arm uses \"arm-none-eabi\" but alternative toolchains may use a different prefix. Currently the gdb and objdump tools are needed."
                 },
@@ -659,6 +662,11 @@
                             "powerOverBMP": {
                                 "type": "string",
                                 "description": "Power up the board over Black Magic Probe. \"powerOverBMP\" : \"enable\" or \"powerOverBMP\" : \"disable\". If not set it will use the last power state."
+                            },
+                            "stopOnEntry": {
+                                "type": "boolean",
+                                "description": "Halt device after debug start",
+                                "default": true
                             }
                         },
                         "required": [
--- a/src/common.ts
+++ b/src/common.ts
@@ -112,6 +112,7 @@ export interface ConfigurationArguments extends DebugProtocol.LaunchRequestArgum
     targetId: string | number;
     cmsisPack: string;
     runToMain: boolean;
+    stopOnEntry: boolean;
 
     // J-Link Specific
     ipAddress: string;
--- a/src/gdb.ts
+++ b/src/gdb.ts
@@ -341,7 +341,15 @@ export class GDBDebugSession extends DebugSession {
                         }, 50);
                     };
 
-                    if (this.args.runToMain) {
+
+                    if (!this.args.stopOnEntry) {
+
+                        setTimeout(() => {
+                            this.stopped = false;
+                            this.miDebugger.sendCommand('exec-continue');
+                        }, 100);
+
+                    } else if (this.args.runToMain) {
                         this.miDebugger.sendCommand('break-insert -t --function main').then(() => {
                             this.miDebugger.once('generic-stopped', launchComplete);
                             // To avoid race conditions between finishing configuration, we should stay
@@ -660,12 +668,21 @@ export class GDBDebugSession extends DebugSession {
 
             this.miDebugger.restart(commands).then((done) => {
                 this.sendResponse(response);
-                setTimeout(() => {
-                    this.stopped = true;
-                    this.stoppedReason = 'restart';
-                    this.sendEvent(new ContinuedEvent(this.currentThreadId, true));
-                    this.sendEvent(new StoppedEvent('restart', this.currentThreadId, true));
-                }, 50);
+
+                if(this.args.stopOnEntry) {
+                    setTimeout(() => {
+                        this.stopped = true;
+                        this.stoppedReason = 'restart';
+                        this.sendEvent(new ContinuedEvent(this.currentThreadId, true));
+                        this.sendEvent(new StoppedEvent('restart', this.currentThreadId, true));
+                    }, 50);
+                } else {
+                    setTimeout(() => {
+                        this.stopped = false;
+                        this.miDebugger.sendCommand('exec-continue');
+                    }, 100);
+                }
+
             }, (msg) => {
                 this.sendErrorResponse(response, 6, `Could not restart: ${msg}`);
             });


haneefdm:
I am not arguing with the functionality (yet) but the nomenclature is bit odd.

In VSCode world [`stopAtEntry`](https://github.com/Microsoft/vscode-cpptools/blob/master/launch.md) means the same as `runToMain` for Cortex-Debug. With cpptols, it does not stop at crt0 startup but actually main. crt0 is the equivalent to reset-vector but that is not what they mean.

Wondering about your definition of 'Entry'? If it is the reset vector, then both `stopAtEntry`  and `runToMain` have to be true` to get the program to break in main? Isn't that confusing? stopAtEntry does not stopAtEntry? (I know you used the word stopOnEntry)

Maybe I am misunderstanding the logic but this is what I see
image. Not calling `launchComplete` may be okay but...

Another potential problem is that you are hoping that the debugger will stop at some breakpoint you set. Well, since you continue immediately, we may not even get a chance to set those breakpoints. If you want to continue, then it is safer to do it after the debugger configuration is completed -- like we do it when `runToMain` is set to true.

kaegi:
Thanks for your feedback! I'm not very proficient with this codebase, so I tried the "code, test and pray" approach (with medium success it seems). Using

```
this.onConfigDone.once('done', () => {
    this.miDebugger.sendCommand('exec-continue');
});
```

did not seem to give the expected results in my short tests, so I left it out. It seems like a good idea to wait for the configuration...

The debugger should not be stopped (like in `launchComplete`), because that would focus the first line in the program for a short time. This is exactly what this change is supposed to suppress (so one can code uninterrupted).

Concerning the names and `runToMain`: If `stopOnEntry` works as expected (=respecting user defined breakpoints) then `runToMain` is a special case that can be deleted. Yes, better names and different flow should be used if both configurations are allowed. What is your opinion on deleting `runToMain`?

I will test the changes again with `this.onConfigDone.once(...)` in the near future.

kaegi:
Clarification: With my setup, `stopAtEntry=true` stops here, which is very distracting:

Screenshot

The `main` function gets called a few lines after that.

haneefdm:
I think not calling `launchComplete` may be okay, just that things like the register/svd views do not get a chance to initialize themselves but they will do it later when the program finally stops.

If doing a `exec-continue` when config. is complete is not working, then there must be something else wrong. Does your callback ever get called? It has to!?!?

I think we will have thousands of irritated users if we remove `runToMain` :-)

kaegi:
The new versions handles both flags as if they were `stopOnEntry` and `stopOnMain` (not exactly what the names suggest but a little bit better).

Using the `onConfig` callback works as expected. Sorry for the false alarm.

```
I think we will have thousands of irritated users if we remove runToMain :-)
```

Good point :)

haneefdm:
One thing though. stopping by default at the reset-vector is normal and expected and most (if not all) embedded gdb servers are designed to do that.

Taking an explicit action to disable it is preferable because the user always knows that you can debug pre-main stuff is possible. I believe other embedded debuggers do something very similar to Cortex-Debug.

kaegi:
If stopping at the reset vector is expected - so be it!

On a side note: I cannot use `runToMain` as a normal configuration (no effect). Is this a bug or am I doing something wrong?
image

haneefdm:
Hmmm. Check your `package.json` changes very carefully. Make sure you are editing the launch section and not the attach section.

And, `runToMain` does not make sense for an `attach` type session. Are you sure you were editing a `launch` type in `launch.json`

I am still uncomfortable about the naming and the truth table I mentioned above. @Marus has the final word though.

kaegi:
Very interesting! I was using the `attach` type. Everything was working as expected nonetheless. For completeness here is my full (corrected) launch configuration:

```
{
    // Debug settings: https://github.com/WebFreak001/code-debug
    "version": "0.2.0",
    "configurations": [
        {
            //  Debug the STM32 Blue Pill program.
            "name": "Debug (STM32 Blue Pill Rust)",
            "type": "cortex-debug",
            "request": "launch",
            "preLaunchTask": "🔨 Build (STM32 Blue Pill Rust)",
            "cwd": "${workspaceRoot}",
            //  This target setting is mandatory but redundant because it's specified in loader.gdb.
            "servertype": "openocd",
            "device": "STM32F103C8",
            "executable": "./target/thumbv7m-none-eabi/debug/stm32-blue-pill-rust",
            "stopOnEntry": false,
            "runToMain": false,
            "configFiles": [
                "interface/stlink-v2.cfg",
                "target/stm32f1x.cfg"
            ],
            "preLaunchCommands": [
                "monitor arm semihosting enable",
            ],
            "overrideLaunchCommands": [
                "load",
                "monitor reset halt",
            ],
            "overrideRestartCommands": [
                "load",
                "monitor reset halt",
            ],
        }
    ]
}
```

haneefdm:
Oh yeah, I forgot to mention that. That timeout would have created very hard to debug race conditions between configuration and the continue. That was happening at one time for `runToMain`. In fact your situation is the same as `runToMain`, if you do not actually set a breakpoint right?

kaegi:
```
In fact your situation is the same as runToMain, if you do not actually set a breakpoint right?
```
Yes, that is one way to look at it.

haneefdm:
The `restart` code you added looks fine to me. But now, I am wondering why the timeout existed in the original code?

kaegi:
Cool! Almost all of the 8 cases (`runToMain=true/false`, `stopAtMain=true/false`, `connect`/`restart`) work reliably in my tests. The only cases that are unexpected are `runToMain` with restarts (the breakpoint is deleted after it is hit, so restarts do not stop at main) which is a separate bug(???). 

```
But now, I am wondering why the timeout existed in the original code?
```
The `connect` code needed the timeout as a replacement for the `onConfig` "timeout" because some breakpoints were not hit at the beginning. I thought the same thing might happen with restarts.

----------------

I think the only case that needs clarification is `runToMain=true, stopOnEntry=true`. Currently it stops on both points. To maintain backwards compatibility, it should behave like `runToMain=true, stopOnEntry=false` (which means to ignore `stopOnEntry` when `runToMain=true`).

On the other hand, this is unexpected, because `stopOnEntry=true` does not stop on entry. I can't think of a simple name that reflects that `stopOnEntry` has a lower priority than `runToMain`.

`stopOnMain` and `stopOnEntry` seem like best names to me (but again - breaking backwards compatibility).

haneefdm:
`runToMain` not being implemented for a restart is intentional. Note that re-start is the same for launch and attach and the attach config. does not even allow runToMain.

haneefdm:
May I ask why you do a `load` for a restart in launch.json?

kaegi:
This way the current executable is flashed onto the chip. This allows for rapid prototyping. This is my workflow:

[Debugger is running]
Changing the source code
Saving with Ctrl+S
Alt+Shift+F5 restarts the debugger
"preLaunchTask" builds the firmware with the changes
"restart->"load" automatically flashes the current firmware with the changes
"restart->"monitor reset halt" resets the device
with "stopOnEntry=false" the "exec-continue" the code runs without stopping...
...maybe until my first breakpoint (probably where I changed the code)
[Debugger is running]
Changing the source code
Saving with Ctrl+S
....

The whole process takes 1s-2s on my machine and does not require more manual steps than necessary (auto-save on debug restart would even take one step away) .

I'm _very_ new to to programming embedded ARM (STM32) and currently only throwing together a template I can use to quickly start new projects (I've only used ESP32 and AVR before, which I feel are a little bit easier due to abstractions). Being able to rapidly prototype always makes understanding the library, hardware, etc. much easier (and less annoying to do the same steps/commands every time I want to change something).

How does your edit-compile-flash-debug configuration look like? And what are "restarts" for you?

haneefdm:
Sounds good. For me, if I have to restart its because I somehow missed an opportunity to examine something and it only happens at the start or debugging the bootloader/startup code. Normally, my programs have a bunch of tasks/events/interrupts and they run forever so no need to restart. I also *try* not to edit while I am debugging as breakpoints can shift and I lose track.

The entire idea is to not reload a program for a restart. For me, a reload can take about 10 seconds (erase, program verify). With cpptools, they literally do an entire start re-launch the gdb+server, etc. In Cortex-Debug, we use the same gdb/server processes.

haneefdm:
I propose implementing the feature described in Issue #197 instead of this feature. It would be more generic.

Marus:
Sorry - had been gone a bit longer than I expected and was swamped with other things when I got back so hadn't got to looking at this until now. I agree with @haneefdm's assessment - I feel like #197 is a more generic system that would cover a lot of these use cases. And again - I don't like like the re-flashing on restart - restart in cortex-debug has always been quite literally resetting the target core.

haneefdm:
@kaegi Please checkout the 0.3.3 release. You should be able to achieve what you want using the new launch.json options `postStartSessionCommands` and `postRestartSessionCommands`. See changelog. Please give us feedback. If that works for you we should close this PR

haneefdm:
@kaegi Also, thank you for your submission. It is just that when I saw your need and a couple of other different needs, we came up with a more generic solution. Hope you understand

kaegi:
Sorry for my inactivity, I'm currently not doing any ARM development. The generic solution sounds nice! I will test it some time in the future.

PR #233, opened 2020-01-23
Fix issue: Unable to view Static variables during debug

Frzzen:
There are two issues that lead to this situation:

In my case,  This function blew not doing the right job
``` typescript
public getStaticVariables(file: string): SymbolInformation[] {
    return this.symbols.filter((s) => s.type === SymbolType.Object && s.scope === SymbolScope.Local && !s.name.includes(".") && (s.file === file || file.endsWith(`/${s.file}`)));
}
``` 

First: The arg (file: string) format like `<file path>/<filename>`  and the s.file is just `<filename>`.

Second: Static local variables are compiled into `<var name>.<hash>` such as `foo.12521`,  The function `getStaticVariables` also include static local variable with error name in result list.

### DIFF
--- a/src/backend/symbols.ts
+++ b/src/backend/symbols.ts
@@ -89,7 +89,7 @@ export class SymbolTable {
     }
 
     public getStaticVariables(file: string): SymbolInformation[] {
-        return this.symbols.filter((s) => s.type === SymbolType.Object && s.scope === SymbolScope.Local && s.file === file);
+        return this.symbols.filter((s) => s.type === SymbolType.Object && s.scope === SymbolScope.Local && !s.name.includes(".") && (s.file === file || file.endsWith(`/${s.file}`)));
     }
 
     public getFunctionByName(name: string, file?: string): SymbolInformation {


src/backend/symbols.ts
```diff
@@ -89,7 +89,7 @@ export class SymbolTable {
     }
 
     public getStaticVariables(file: string): SymbolInformation[] {
-        return this.symbols.filter((s) => s.type === SymbolType.Object && s.scope === SymbolScope.Local && s.file === file);
+        return this.symbols.filter((s) => s.type === SymbolType.Object && s.scope === SymbolScope.Local && !s.name.includes(".") && (s.file === file || file.endsWith(`/${s.file}`)));
```
haneefdm: So, if I have a duplicate file name like `myfile.c` in two different directories, we could get a match for both. I am okay with that but it could be confusing if/when it really happens. Do we know the root cause of why this is even happening?

Frzzen:
> So, if I have a duplicate file name like myfile.c in two different directories, we could get a match for both. 

Sorry, I didn't notice that, I share  why does this issue appear below, but do not have a good idea to fix it yet.

This method def in [mi2.ts](https://github.com/Marus/cortex-debug/blob/master/src/backend/mi2/mi2.ts) to get the file info:
``` ts
public getFrame(thread: number, frame: number): Thenable<Stack> {
        return new Promise((resolve, reject) => {
            const command = `stack-info-frame --thread ${thread} --frame ${frame}`;

            this.sendCommand(command).then((result) => {
                const frame = result.result('frame');
                const level = MINode.valueOf(frame, 'level');
                const addr = MINode.valueOf(frame, 'addr');
                const func = MINode.valueOf(frame, 'func');
                const file = MINode.valueOf(frame, 'file');
                const fullname = MINode.valueOf(frame, 'fullname');
                let line = 0;
                const linestr = MINode.valueOf(frame, 'line');
                if (linestr) { line = parseInt(linestr); }

                resolve({
                    address: addr,
                    fileName: file,
                    file: fullname,
                    function: func,
                    level: level,
                    line: line
                });
            }, reject);
        });
    }
``` 

And in [symbols.ts](https://github.com/Marus/cortex-debug/blob/master/src/backend/symbols.ts) you can find this
``` ts
public loadSymbols() {
        try {
            let objdumpExePath = os.platform() !== 'win32' ? `${this.toolchainPrefix}-objdump` : `${this.toolchainPrefix}-objdump.exe`;
            if (this.toolchainPath) {
                objdumpExePath = path.normalize(path.join(this.toolchainPath, objdumpExePath));
            }

            const objdump = childProcess.spawnSync(objdumpExePath, ['--syms', this.executable]);
            const output = objdump.stdout.toString();
            const lines = output.split('\n');
            let currentFile: string = null;
            
            for (const line of lines) {
                const match = line.match(SYMBOL_REGEX);
                if (match) {
                    if (match[7] === 'd' && match[8] === 'f') {
                        currentFile = match[11].trim();
                    }
                    const type = TYPE_MAP[match[8]];
                    const scope = SCOPE_MAP[match[2]];
                    let name = match[11].trim();
                    let hidden = false;

                    if (name.startsWith('.hidden')) {
                        name = name.substring(7).trim();
                        hidden = true;
                    }

                    this.symbols.push({
                        address: parseInt(match[1], 16),
                        type: type,
                        scope: scope,
                        section: match[9].trim(),
                        length: parseInt(match[10], 16),
                        name: name,
                        file: scope === SymbolScope.Local ? currentFile : null,
                        instructions: null,
                        hidden: hidden
                    });
                }
            }
        }
        catch (e) { }
    }
```

So, the result is, when use GDB/MI call the `-stack-info-frame` command in [mi2.ts](https://github.com/Marus/cortex-debug/blob/master/src/backend/mi2/mi2.ts),  will get file path and filename, but with `objdump`in  [symbols.ts](https://github.com/Marus/cortex-debug/blob/master/src/backend/symbols.ts) just get the `file name !
  
截屏2020-01-23下午7 11 31

So, if we can get the every same files' path, this problem will be solved, 
I try use gcc which xcode support every thing is OK, I can find the path above the file name:
```
$ gcc -v
Configured with: --prefix=/Applications/Xcode.app/Contents/Developer/usr --with-gxx-include-dir=/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/include/c++/4.2.1
Apple clang version 11.0.0 (clang-1100.0.33.16)
Target: x86_64-apple-darwin19.2.0
Thread model: posix
InstalledDir: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin

$ objdump --syms a.out

a.out:  file format Mach-O 64-bit x86-64

SYMBOL TABLE:
0000000000000000 l    d  *UND*  /Users/frzzen/Projects/test_objdump_path/
0000000000000000 l    d  *UND*  main.c
000000005e2976e2 l    d  __DATA,__data  __data
0000000100000fa0 l    d  __TEXT,__text  __text
0000000100000fa0 l    d  __TEXT,__text  __text
0000000000000016 l    d  *UND*
0000000000000016 l    d  __TEXT,__text  __text
0000000000000000 l    d  __TEXT,__text  __text
0000000000000000 l    d  *UND*  /Users/frzzen/Projects/test_objdump_path/children/
0000000000000000 l    d  *UND*  main.c
000000005e2976e2 l    d  __DATA,__data  __data
0000000000000000 l    d  *UND*  _foo
0000000000000000 l    d  __TEXT,__text  __text
0000000100000000 g     F __TEXT,__text  __mh_execute_header
0000000100001000 g     O __DATA,__data  _foo
0000000100000fa0 g     F __TEXT,__text  _main
0000000000000000         *UND*  dyld_stub_binder
```
but with arm-none-eabi-gcc, things get weird :
```
$ arm-none-eabi-gcc --version
arm-none-eabi-gcc (GNU Tools for Arm Embedded Processors 9-2019-q4-major) 9.2.1 20191025 (release) [ARM/arm-9-branch revision 277599]
Copyright (C) 2019 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.

$ arm-none-eabi-objdump -t build/out.elf
00000000 l    df *ABS*  00000000 gpio.c
00000000 l    df *ABS*  00000000 stm32f1xx_it.c
00000000 l    df *ABS*  00000000 stm32f1xx_hal_msp.c
00000000 l    df *ABS*  00000000 stm32f1xx_hal.c
00000000 l    df *ABS*  00000000 stm32f1xx_hal_rcc.c
08000318 l     F .text  0000002c RCC_Delay
00000000 l    df *ABS*  00000000 stm32f1xx_hal_cortex.c
00000000 l    df *ABS*  00000000 system_stm32f1xx.c
00000000 l    df *ABS*  00000000 init.c
00000000 l    df *ABS*  00000000 main.c
00000000 l    df *ABS*  00000000 /opt/gcc-arm-none-eabi/gcc-arm-none-eabi-9-2019-q4-major/bin/../lib/gcc/arm-none-eabi/9.2.1/thumb/v7-m/nofp/crti.o
00000000 l    df *ABS*  00000000 /opt/gcc-arm-none-eabi/gcc-arm-none-eabi-9-2019-q4-major/bin/../lib/gcc/arm-none-eabi/9.2.1/thumb/v7-m/nofp/crtn.o
00000000 l    df *ABS*  00000000 main.c
```
I'm stuck here now 😢

haneefdm:
Thanks @Frzzen . I have a NUCLEO-F429ZI board. Could you help me duplicate this problem? I may need a small sample project and the exact version of gcc (a link to the download would be good) you are using and the OS.

Frzzen:
Hi @haneefdm 
FYI
gcc:
[arm-none-eabi-gcc version: 9-2019-q4-major](https://developer.arm.com/tools-and-software/open-source-software/developer-tools/gnu-toolchain/gnu-rm/downloads)
os: 
截屏2020-01-23下午11 41 26
I create the project by use STM32CubeMX and my chip is STM32F103C8T6, I push the project [Here](https://github.com/Frzzen/issue-example-pr233)(I tested,  same problem arise).
And for your board I create a [sample project Here](https://github.com/Frzzen/issue_example_pr233_stm32f429) but I have not the same board right now, so I do not know does this work well or not.

I hope it helps

haneefdm:
Thanks, I will look at it soon. Chasing another bug right now :-)

haneefdm:
Okay, that download of the GNU tools was painful on macOS Catalina. I had to individually approve/sign every executable. One thing missing in the toolchain though. **OpenOCD**. Not that it matters but I want to use a version that works because there are SO many versions out there. If you have a zip file of your OpenOCD build for macOS, that would work fine.

In the meantime, I will try to get **a** version. The version I have is super vendor-specific.

haneefdm:
Whew!!. That took a while but I am able to duplicate the problem using STLink config. Thanks for providing this nice/simple test case. The `Local` section in Variables Window is not listing the `static_local_var` but the watch window does. That is the gist of it right?

Let's see if I can make actual progress on it. Onwards.

Frzzen:
I'm glad it works! @haneefdm 
But it is a bit different in my case, I can not find the `static_var` in the `VARIABLES/Static` , but `WATCH` does:  
截屏2020-01-26上午11 21 38

And, the OpenOCD  I used is [here](https://github.com/xpack-dev-tools/openocd-xpack/releases/download/v0.10.0-13/xpack-openocd-0.10.0-13-darwin-x64.tgz)

I'm glad to be able to help 😊

haneefdm:
@Frzzen I just pushed changes (to my repo) to better support static variables. It was a bit more involved and certain other things did not work wrt. statics. If you have the time, can you check it out to make sure it addresses your case?

https://github.com/haneefdm/cortex-debug/commit/ee3f458cc24e44a5fd2a36de0238a881134edf11

I will do a PR after some more testing and a self-code review. You are welcome to review changes as well.

Issue #232, opened 2020-01-22
No SVD File Loaded


dzarda:
The SVD view shows the "No SVD File Loaded" message. This appears to be a recurring issue judging by the recently closed issues with the same problem. Tried relative and absolute paths in the SVD path, tried forward slashes:

```json
{
    "version": "0.2.0",
    "configurations": [
        {
            "type": "cortex-debug",
            "request": "launch",
            "servertype": "openocd",
            "svdFile": "C:/Users/Dzarda/Documents/src/lol/stm32l0x3.svd",
            "cwd": "${workspaceRoot}",
            "executable": "./target/thumbv6m-none-eabi/debug/lol",
            "name": "Debug (OpenOCD)",
            "device": "STM32L053",
            "configFiles": [
                "interface/stlink.cfg",
                "target/stm32l0.cfg",
            ]
        }
    ]
}

```

Happens in a Rust project.

Using [SVD](https://stm32.agg.io/rs/stm32l0x3.svd.patched) from this project: https://github.com/stm32-rs/stm32-rs

haneefdm:
Maybe I can help. But a bit more context, please?

What recurring issue? What do you see? Assuming an issue with SVD, but what are you observing?

dzarda:
Please excuse me, updated the description. Not sure about reproducability.

haneefdm:
Hi @dzarda, I doubt this is a rust specific issue. Cortex-Debug doesn't even know what rust is. I could not figure out **exactly** which SVD file is an issue. Sometimes, this can happen if the SVD is un-parseable or there is a bug in the SVD parser. Please point to the SVD file if you can.

dzarda:
SVD link is in the description, also here: https://stm32.agg.io/rs/stm32l0x3.svd.patched

haneefdm:
Sorry, I did not notice the SVD link earlier. I can now duplicate the problem. There was a problem with the DMA1 peripheral (around line 756) and resulted in the SVD file parsing getting aborted.

Not sure why and investigating. Thanks.

haneefdm:
Okay, I found and fixed the issue. It had to do with a `derivedFrom` attribute on enum values. It was not supported/expected by Cortex-Debug. However, even the current spec is unclear as to how it **should** be interpreted.

https://www.keil.com/pack/doc/CMSIS/SVD/html/elem_registers.html#elem_enumeratedValues

[...]

PR #234, opened 2020-01-24
Fix Issue#232

haneefdm:


haneefdm:
This is to fix Issue #232 

Maybe we need a better mechanism as well when a parsing causes an exception. I will look into that. Right now, it simply says 'No SVD File Loaded'. With this message, we can't tell if the file was not found or there were other issues. I will do this in another round.

PR #240, opened 2020-01-31
Fix issue #239

### LINKED ISSUE #239: Ability to specify RTT telnet port or arbitrary server args
Hi.

I really like your extension. I think it lacks one major function. I am developing a system where two devices talk to each other, so my setup consists of two simultaneously debugged devices. The thing is - `JLinkGDBServerCL` which is my debug server by default uses fixed port for RTT communication with debugged devices, so with current setup I am able to use RTT only on one of the debuggers (the one that was started first). There is, however `-RTTTelnetPort` switch for the server CLI that forces it to use specified port. Unfortunately I am unable to force my `launch.json` to properly pass that arg when starting server. It would be great to give an option to either:
- specify `RTTTelnetPort` directly with dedicated option in launch.json, or
- give an ability to pass arbitrary args to server (with an option like `serverArgs`, exactly like `debuggerArgs`)

Thanks!

sebastian-bartkowiak:
Fixes #239

### DIFF
--- a/package.json
+++ b/package.json
@@ -279,9 +279,10 @@
                                 "type": "string"
                             },
                             "serverArgs": {
-                                "default": null,
+                                "default": [],
                                 "description": "Additional arguments to pass to gdb-server command line",
-                                "type": "array"
+                                "type": "array",
+                                "items": "string"
                             },
                             "executable": {
                                 "description": "Path of executable",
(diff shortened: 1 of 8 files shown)


package.json
```diff
@@ -278,6 +278,11 @@
                                 "description": "This setting can be used to override the gdb-server path user/workspace setting for a particular launch configuration. It is the full pathname to the executable or name of executable if it is in your PATH",
                                 "type": "string"
                             },
+                            "serverArgs": {
+                                "default": null,
```
haneefdm: Just like the items that are an array of string items, you might want to
```JSON
                                "default": [],
                                "type": "array",
                                "items": "string",
```
to get better JSON validation in the editor.

PR #245, opened 2020-02-09
Serialport v12.4.0

haneefdm:
Hi Marcel,

This PR contains the script and updated serialport for all 64-bit platforms. I did not add usb because it was failing on linux. I can give a another try and do another PR. Oh also contains a unit test just to see if the appropriate version exists and if the serialport module can be loaded into the extension

We could delete the old serialport I think but thought you could decide on that later. Let me know if you want me to do it on this same branch, maybe after this PR, I can do another.

Thank you Marcel, for your help


Marus:
Looks good.

PR #246, opened 2020-02-09
Webpack - Binary Module Loading Fix

Marus:
This should fix loading of the binary modules for the serial port when webpack has been completed. Webpack normally replaces the require method - but you can access it using the _non_webpack_require__ method (which becomes regular require when packed). This change conditionally uses with require or __non_webpack_require__ depending on if it is being used in a webpack context.

### DIFF
--- a/src/frontend/swo/sources/serial.ts
+++ b/src/frontend/swo/sources/serial.ts
@@ -5,6 +5,9 @@ import * as os from 'os';
 import * as vscode from 'vscode';
 import * as path from 'path';
 
+declare function __webpack_require__(name: string): any;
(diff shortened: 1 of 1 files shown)


haneefdm:
Ohh. I never saw this issue bother me. I may have to update the unit test

Marus:
Yeah - seems to cause an issue with newer versions of webpack I expect - but could definitely see when I looked at the compiled source that it was using the wrong require currently. 

In the case of the tests it won't see that though - as the tests aren't run through webpack. So while they test that the binary modules are up-to-date and loadable (at least on the platform the tests are run against) - they wouldn't pick up this "web-packing" issue.

PR #248, opened 2020-02-11
Significant changes and updates, bug fixes

haneefdm:
I updated the ChangeLog to show externally visible changes. Please edit the version number in the ChangeLog when you decide to merge. I wasn't sure how to name it as of now in package.json.

Internal changes:
* When we invoke `objdump`, we now try to get more debug info and this helps in figuring out symbols origin. Not perfect but much than before when statics were not working at all unless path names matched exactly between `gdb` and `objdump`. They don't. `objdump` only gives you the base name. `gdb` world view is not possible to duplicate even with `MI3` protocol which ARM doesn't support yet
* we better present anonymous unions/structs, as they were getting lost in VSCode, protocol and there were missing union/structs. But now we can see everything, preserve our original presentation method as well as what other debuggers (Eclipse, Visual Studio) use.
* enabled an option for de-mangling. But this had a ripple effect as we used `::` as a separator between file and symbol but `::` is a natural separator in C++/Rust for ordinary symbols. I changed `::` to `:::` between file and symbol.
* modified how we maintain globals and statics and children of certain scopes. This now fixes issues with setting static variable because given a reference and a simple/display name, we can figure out what the hell it is.
* a few minor defensive coding changes.

In this PR, I have no updates related to the serialport loading issues.


haneefdm:
One or two items in my ChangeLog are actually changes contributed by other PRs. Just consolidated them

haneefdm:
Sorry my last commit  0fa7486 should have said it solves Issue #249

PR #251, opened 2020-02-21
Remove webpack processing for extension/debug adapter to avoid issues with loading binary modules 

Marus:


Marus:
Going to close this one and instead go for the webpack fix I think.

PR #294, opened 2020-05-19
add resetAfterDownload for jlink.

jkent:
This is useful for targets that perform hardware DMA to RAM upon reset.  This is the case for the POLLUX VR3520F/LF1000


jkent:
Woops, I just learned about overrideLaunchCommands!  Closing this.

haneefdm:
It was also a breaking change and an incomplete one. Thanks for closing the
PR.

PR #302, opened 2020-06-11
not working if rtos specified

kennethzhu:


haneefdm:
@kennethzhu Actually, your changes make lot of sense to me. Applies even if there is no RTOS

But what was not working when RTOS enabled? Can you provide more details? Which RTOS? Which gdb-server, device etc. So I can test. Perhaps a launch.json included? Is there an issue that you file already that this fixes? I can go look at that.

haneefdm:
Can you try https://github.com/haneefdm/cortex-debug/releases/tag/v0.3.7-pre3

haneefdm:
Your changes and one more like it was merged manually by me. So, closing this PR

Issue #284, opened 2020-04-30
SVD File not loading the derivedFrom values


danielhep:
My SVD file makes heavy use of the "derivedFrom" attribute on some registers, since many registers share the same set of fields. However, in cortex-debug, the fields only show up on the registers on which they are defined, and not those which are derived from them. 

Here's an example of the register that uses derivedFrom:
```
                <register derivedFrom="P300PFS">
                    <dim>10</dim>
                    <dimIncrement>0x4</dimIncrement>
                    <dimIndex>0-9</dimIndex>
                    <name>P00%sPFS</name>
                    <description>P00%s Pin Function Control Register</description>
                    <addressOffset>0x000</addressOffset>
                    <size>32</size>
                    <access>read-write</access>
                    <resetValue>0x00000C00</resetValue>
                    <resetMask>0xFFFFFFFD</resetMask>
                </register>
```
The P300PFS register has a `<fields>` tag on it that I would expect to carry forward to the registers that result from the above, however as you can see in this screenshot, only the "parent" registers have any fields on them. 

![image](https://user-images.githubusercontent.com/631230/80769545-6346c800-8b02-11ea-9309-8044bffceebe.png)

Based on what I read [here](https://www.keil.com/pack/doc/CMSIS/SVD/html/elem_registers.html#elem_register), it seems like this syntax is correct, since both registers are located in the same peripheral, so it should not be necessary to use the peripheral.register notation. If this is a problem in cortex-debug, feel free to point me in the right direction for fixing it, and I can submit a pull request.

haneefdm:
Is it possible to have the full svd file?

haneefdm:
Can you provide more info, please? Debugging SVD file issues is not easy. Please respond @danielhep I need the actual SVD file.

I am about to make a new release and hope to get your issue resolved before then. I am sure you have a real issue but I need to be able to reproduce it.

danielhep:
@haneefdm I sent you the SVD via email. Let me know if that works for you. Thank you!

haneefdm:
Could you please try the following pre-release

https://github.com/haneefdm/cortex-debug/releases/tag/v0.3.7-pre3

Use the VSIX manual installation method to install the extension

danielhep:
It looks like the registers are showing up, but they are not showing accurate information. Also, writing to them doesn't seem to work. These registers control an IO connected to an LED, so that is what I'm testing with. 

[...]

Issue #273, opened 2020-04-14
Parsing issue of SVD file


josko7452:
Hi I have trouble getting SVD file loaded from:
https://github.com/m-labs/dslite2svd/blob/master/svd/tm4c123x.xml

I think it will be likely related to #232 or it is a different parsing error.

haneefdm:
Could you post the error(s) you are seeing?

josko7452:
It's just No SVD File Loaded <PATH>
Where <PATH> is correct path to my file (i.e. E:\WORK\Data\tm4c123x.svd)

haneefdm:
Found the issue, fix is coming. We were not expecting peripherals without an `addressBlock`, and "_INTERRUPTS" was one such peripheral.

haneefdm:
Could you please try the following pre-release

https://github.com/haneefdm/cortex-debug/releases/tag/v0.3.7-pre3

Use the VSIX manual installation method to install the extension

josko7452:
@haneefdm It works on the file I've linked above.

PR #303, opened 2020-06-14
Preparation for Version 0.3.7

haneefdm:
From the ChangeLog...

Minor bug fix release

1. New feature
   * Support for IAR produced elf files which just did not work with V0.3.6. They didn't quite work with earlier releases either. They are non-standard (kinda) compared to what gcc produced or what objdump documents
   * With the latest version of VSCode, there were too many popups coming. They are all harmless, but annoying. There were changed made to suppress these but they caused other problems. We may just have to live with those popups (that are non-modal and disappear after a while) until we figure out how VSCode has changed. May take a while
2. Issues Fixed
   * Fixed Issue [#273](https://github.com/Marus/cortex-debug/issues/273). Handle peripherals blocks with no addressBlocks. Also fix issue with multiple address blocks where only the first one was being used
   * Fixed Issue [#284](https://github.com/Marus/cortex-debug/issues/284). Implement/support 'derivedFrom' at register level.
   * Fixed Issue [#306](https://github.com/Marus/cortex-debug/issues/306). Support displaying more than 20 stack frames
   * When runToMain was enabled there were 1-2 harmless popups when the program stopped in main. They were very frequent on Windows, less frequent on Linux and very rare if any on a Mac.


haneefdm:
A preview of this release is available at

https://github.com/haneefdm/cortex-debug/releases

PR #304, opened 2020-06-16
Bugfix: Add a delay before reading the SWO file

GrandChris:
On debugging on a STM32F765NG on Windows, OpenOCD needs too much time to create the SWO file (probably because of a long configure Script). There needs to be more time added before Cortex Debug tries to read from the SWO file. Otherwise the error message "Unable to open path: ${SWOPath} - Unable to read SWO data." happens.

### DIFF
--- a/src/frontend/swo/sources/file.ts
+++ b/src/frontend/swo/sources/file.ts
@@ -3,6 +3,15 @@ import * as vscode from 'vscode';
 import { EventEmitter } from 'events';
 import { SWOSource } from './common';
 
+
+function sleep(milliseconds) {
+    const date = Date.now();
+    let currentDate = null;
+    do {
+      currentDate = Date.now();
+    } while (currentDate - date < milliseconds);
+  }
+
 export class FileSWOSource extends EventEmitter implements SWOSource {
     public connected: boolean = false;
     private fd: number = null;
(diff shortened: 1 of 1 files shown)


haneefdm:
@GrandChris There may be a better way of doing this than sleeping in a busy loop.

Could you please file an issue with your sample launch.json and the output of OpenOCD/DebugConsole please? WIth `"showDevDebugOutput: true" in launch.json please. We can track progress this way and have a specific problem we are debugging/fixing

I am asking because we already wait for a message from OpenOCD that it is ready before we try to open that file or even start gdb.

Waiting for 500ms is a very long time as well. What was the minimum sleep time you found that worked for you? I want to understand the problem better. 

It would also be good to see your openocd config file(s)/command(s). Reason is that perhaps you are doing a reset/init too soon before configuring.

With your help, I want to get this into the next release, hopefully this weekend.

GrandChris:
I created a new Issue "Can't read SWO data with OpenOCD on a STM32F765NG #305"
Since this is only a one line fix, this merge request is more a conversation base. I also do not think that busy waiting a fixed time is a good idea. I'm open for any suggestions. Maybe it is possible to watch the filesystem, or get reasonable feedback from OpenOCD. 
While the implementation is questionable, it does however solve this problem.

Open the SWO file fails with a sleep time of 50ms and works with a sleep time of 75ms. I just like to be on the safe side.

haneefdm:
Sorry, I just made a release without changes to address this issue.

I needed some testing on #305. But, there were many other things that were waiting to be pushed out, so I had to go ahead.

Catch you on the next release

PR #311, opened 2020-07-07
Eight Bit Register support & Refresh on update peripheral register

danielhep:
We noticed that when loading values from 8 bit registers on our MCU, the values were being displayed incorrectly, or in the wrong register. We believe this to be an endianess issue. The fix was to send commands to load each register individually instead of entire peripherals at a time. It is less efficient but it produces correct results. Debugging in eclipse had the same bug for us, so this was a big improvement. 

The other change was to wait for updateData() to finish before telling VSCode to update the peripheral tree. This fixes a refresh issue with the tree when updating values, where the values being displayed were not the latest values after changing one. 

Our changes are probably not super well integrated with the rest of the code because we are not super familiar with the code. I'm open to feedback and happy to improve or fix things to make it work better with the rest of the code.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
     "name": "cortex-debug",
-    "version": "0.3.6",
+    "version": "0.3.7",
     "lockfileVersion": 1,
     "requires": true,
     "dependencies": {
--- a/src/frontend/views/nodes/peripheralclusternode.ts
+++ b/src/frontend/views/nodes/peripheralclusternode.ts
@@ -25,7 +25,7 @@ export class PeripheralClusterNode extends PeripheralBaseNode {
     public readonly resetValue: number;
     public readonly accessType: AccessType;
     public currentValue: number;
-    public tempValue: number[];
+    public currentValueArray: number[];
 
     constructor(public parent: PeripheralNode, options: ClusterOptions) {
         super(parent);
--- a/src/frontend/views/nodes/peripheralclusternode.ts
+++ b/src/frontend/views/nodes/peripheralclusternode.ts
@@ -24,6 +24,8 @@ export class PeripheralClusterNode extends PeripheralBaseNode {
     public readonly size: number;
     public readonly resetValue: number;
     public readonly accessType: AccessType;
+    public currentValue: number;
+    public tempValue: number[];
 
     constructor(public parent: PeripheralNode, options: ClusterOptions) {
         super(parent);
--- a/src/frontend/views/nodes/peripheralnode.ts
+++ b/src/frontend/views/nodes/peripheralnode.ts
@@ -102,27 +102,12 @@ export class PeripheralNode extends PeripheralBaseNode {
         return new Promise((resolve, reject) => {
             if (!this.expanded) { resolve(false); return; }
 
-
             const promises = this.children.map(r => {
-                let promise = MemReadUtils.readMemory(this.baseAddress+r.offset, r.size/8, r.tempValue);
+                let promise = MemReadUtils.readMemory(this.baseAddress+r.offset, r.size/8, r.currentValueArray);
                 promise.then(() => r.updateData(), ()=>{});
                 return promise;
             })
-            return Promise.all(promises).then(() => resolve(true), reject);
-            this.readMemory().then((unused) => {
-                const promises = this.children.map((r) => r.updateData());
-
-                Promise.all(promises).then((_) => {
-                    resolve(true);
-                }).catch((e) => {
-                    const msg = e.message || 'unknown error';
-                    const str = `Failed to update peripheral ${this.name}: ${msg}`;
-                    if (vscode.debug.activeDebugConsole) {
-                        vscode.debug.activeDebugConsole.appendLine(str);
-                    }
-                    reject(new Error(str));
-                });
-            }, (e) => {
+            return Promise.all(promises).then(() => resolve(true)).catch(e => {
                 const msg = e.message || 'unknown error';
                 const str = `Failed to update peripheral ${this.name}: ${msg}`;
                 if (vscode.debug.activeDebugConsole) {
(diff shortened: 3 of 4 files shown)


haneefdm:
Thank you for the PR. I think the entire extension assumes little-endian.

But, I have some basic questions. What does your SVD file look like? Is it big-endian?

If you have an 8-bit register, then we should just be reading that address. A byte read will always be correct right? Please educate us on what is going on.

I will take a look at the refresh issue separately.

For stuff like this, it may be better to submit an Issue first before we attempt a PR.

danielhep:
Hi Haneef,
You should already have my SVD file, I sent it to you a little bit ago via email. Just search for my name in your email. The registers are all little endian. 

Before our modifications, your code seemed to be reading in all the registers in a node/group at a time. For example, when you would expand a group in the register tree, all the registers in that node would be loaded with one GDB request. GDB seems to behave differently when loading 32 bits from when you load <32 bits. We believe that is where this issue is coming from, possibly. The endianness issue is at the byte level, not the bit level. GDB is reversing the bytes when you load 32 bits, but not when you load 8, 16, or 24 bits. You can play around with that in GDB to see what I'm talking about.  

Don't feel that you need to use our PR. We aren't familiar with your code base, but it was worth it for us to spend a few hours fixing it immediately and decided to submit it in case you found it useful.

Marus:
sorry I hadn’t looked at this myself before now; I had been on a bit of a hiatus from the project for a while as I had been too busy with other things to really look at much of anything. 

As Haneef mentioned- currently we largely assume little-endian throughout a lot of the code, as while in theory cortex-m supports both big and little endian, in practice almost every released cortex-m microcontroller uses little endian (to be honest I hadn’t been able to find any examples where it didn’t, although not saying they don’t exist). 

While I’m not opposed to trying to support big-endian ones as well if we have examples of this in the wild, I don’t like the approach taken here with the modifications. There are big drawbacks to the individual memory read request for each register - in fact when I originally built the peripheral  register view I took that approach and changed it because you can get very poor performance as it is not uncommon to find peripherals with dozens (I’ve seen some with approximately 100) - so if you have a few peripherals you are working with you could easily get hundreds of memory requests that had to be processed for each step you executed. 

If we’re going to support this in any way I’m going to need more information. What is the particular MCU (and is there somewhere I can get a sample dev board as this is an unusual configuration- so good to have one to reference if possible)? I’d need the SVD definitions (please attach here - email isn’t a good method for disseminating information in a shared project).

Marus:
@danielhep - realize afterwards that you said you didn't think it was an endianness issue. 

You seem to stating that you seemed to think it was some difference in how GDB handled reading when reading <= 32 bits or > 32 bits at a time - but I'm just not seeing that when I'm trying to experiment here and see if I can find an explanation for these issues.

I just tried issuing a bunch of memory read commands in a little test bed application (it happens to be part of the memory block for a GPIO peripheral in my case - but if the reading was happening differently in the way you suggest it wouldn't matter as I'm just manually issuing the reading commands).

I first tried issuing a read for 8 bytes from the address `0x40010800`, then reading each of the 8 bytes one byte at a time - using the same GDB commands that we'd use for reading the peripheral data in the SVD viewer (in fact we only have one memory reading command implemented - it uses the `data-read-memory-bytes` GDB MI command).

The log is below, as you can see the data comes back in the exact same order, namely `0x40 0x44 0x24 0x42 0x44 0x44 0x84 0x88` 

```
-data-read-memory-bytes "0x40010800" 8
{"token":272,"outOfBandRecord":[],"resultRecords":{"resultClass":"done","results":[["memory",[[["begin","0x40010800"],["offset","0x00000000"],["end","0x40010808"],["contents","4044244244448488"]]]]]}}
-data-read-memory-bytes "0x40010800" 1
{"token":121,"outOfBandRecord":[],"resultRecords":{"resultClass":"done","results":[["memory",[[["begin","0x40010800"],["offset","0x00000000"],["end","0x40010801"],["contents","40"]]]]]}}
-data-read-memory-bytes "0x40010801" 1
{"token":131,"outOfBandRecord":[],"resultRecords":{"resultClass":"done","results":[["memory",[[["begin","0x40010801"],["offset","0x00000000"],["end","0x40010802"],["contents","44"]]]]]}}
-data-read-memory-bytes "0x40010802" 1
{"token":141,"outOfBandRecord":[],"resultRecords":{"resultClass":"done","results":[["memory",[[["begin","0x40010802"],["offset","0x00000000"],["end","0x40010803"],["contents","24"]]]]]}}
-data-read-memory-bytes "0x40010803" 1
{"token":151,"outOfBandRecord":[],"resultRecords":{"resultClass":"done","results":[["memory",[[["begin","0x40010803"],["offset","0x00000000"],["end","0x40010804"],["contents","42"]]]]]}}
-data-read-memory-bytes "0x40010804" 1
{"token":161,"outOfBandRecord":[],"resultRecords":{"resultClass":"done","results":[["memory",[[["begin","0x40010804"],["offset","0x00000000"],["end","0x40010805"],["contents","44"]]]]]}}
-data-read-memory-bytes "0x40010805" 1
{"token":171,"outOfBandRecord":[],"resultRecords":{"resultClass":"done","results":[["memory",[[["begin","0x40010805"],["offset","0x00000000"],["end","0x40010806"],["contents","44"]]]]]}}
-data-read-memory-bytes "0x40010806" 1
{"token":181,"outOfBandRecord":[],"resultRecords":{"resultClass":"done","results":[["memory",[[["begin","0x40010806"],["offset","0x00000000"],["end","0x40010807"],["contents","84"]]]]]}}
-data-read-memory-bytes "0x40010807" 1
{"token":191,"outOfBandRecord":[],"resultRecords":{"resultClass":"done","results":[["memory",[[["begin","0x40010807"],["offset","0x00000000"],["end","0x40010808"],["contents","88"]]]]]}}
```

So I certainly don't look like in general GDB is always doing something strange here and treating >32 bit reads differently from < 32bit reads. 

Without a bit more information about exactly what you were using, if you're still trying to use the extension, I really don't have anything else to go off of.

haneefdm:
I am closing this as I don't see a resolution or even a problem. Please engage and reopen if you wish.

PR #327, opened 2020-08-11
Add support for targetDownload option that can skip target-download GDB command.

tritao:
This can be useful if the debug target is already properly flashed, and prevents GDB from initiating a download procedure to the debug server.

Not skipping this can lead to the wrong binaries being flashed to a board.

### DIFF
--- a/package.json
+++ b/package.json
@@ -900,6 +900,11 @@
                                 "maximum": 9,
                                 "default": 0
                             },
+                            "targetDownload": {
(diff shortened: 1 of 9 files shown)


Marus:
This is the intended difference between a `attach` request type and a `launch` request type in your launch.json.

Attach should attach the debugger to an already flashed and running target - while "launch" will run it from scratch (i.e. flash the code to the device and start debugging).

Can you try out using the `attach` request type - I expect this will fix your issue.

Issue #308, opened 2020-06-23
Connect gdb to ip instead of localhost to debug from WSL2


carlolars:
Hi

I'm trying to debug from VSCode running remotely in WSL2 using OpenOCD, which needs to run on the Windows side.

The plugin successfully starts openocd on the windows side, I just added `"openOCDPreConfigLaunchCommands": [ "bindto 0.0.0.0" ]` so that it also listens to the WSL network interface and not only 127.0.0.1, and I have verified that it works by manually starting gdb in WSL2 and connecting to openocd started by the plugin using `<hostip>:50000`.

But how do I configure the plugin so that gdb connects to `<hostip>:50000` instead of `localhost:50000`?

haneefdm:
You can use whatever server you want with an external server.
```JSON
            "servertype": "external",
            "gdbTarget": "host:NNNN",
```
Use your own host (any hostname or ip-addr accessible from your device) and port$. Be sure to start the server first and then launch the debugger.

carlolars:
Alright that works! Not what I'd hoped for since it introdues an additional step but I can live with it.

Here is how I made it work with VSCode running remotely in WSL2:

## Cortex-Debug in WSL2
1. Make sure that `openocd.exe` is found in WSL2 (`.../OpenOCD/0.10.0-13/bin` added to PATH) and that it is executable. 
2. Create a VSCode task that starts OpenOCD and configures it to listen to all interfaces (`"bindto 0.0.0.0"`), see _tasks.json_ below.
3. Create a VSCode launch configuration that connects to the OpenOCD gdb-server running in Windows-land, see _launch.json_ below.
    * The IP address in the `gdbTarget` (172.31.240.1) is the IP address of the Windows host. It can easily be found by in WSL2 do `cat /etc/reslov.conf` where the _nameserver_ IP is the IP of the Windows host. The IP will change when rebooting Windows!
4. Before starting a debug session make sure that the OpenOCD task is running. The task will stay until stopped or VSCode is closed so it usually only has to be started once.

### tasks.json
```
{
    // See https://go.microsoft.com/fwlink/?LinkId=733558
    // for the documentation about the tasks.json format
    "version": "2.0.0",
    "tasks": [
        {
            "label": "Launch OpenOCD",
            "command": "openocd.exe",
            "args": [
                "-f",
                "${input:openocd.config}",
                "-c",
                "\"bindto 0.0.0.0\""
            ],
            "type": "shell",
            "isBackground": true,
[...]

PR #328, opened 2020-08-13
Add support for remote target ip in GDB for WSL2 environments

tamisoft:
The reasoning behind this, is using wsl2 in windows has no access
to the usb port, so one has to use jlinkgdbserver on the host, however
due to some changes in the WSL2, localhost no longer has access to the
host's listening ports, but instead the "default gateway" address
has to be used for connections. Though that is changing on reboot.

Without killing the gdb server between debug sessions, weird side effects
can occur (especially on exceptions) -> normally it gets stuck, and
restarting the gdbserver process is needed.

In this WSL2 scenario, one can use serverArgs -noLocalhostOnly to allow
remote access to the GDB server. And start the native windows executable
for JlinkGDBServerCL.exe in serverpath.

A better approach could be to introduce a wsl2 bool, which could
automatically get the default gw IP address and populate these two variables.

### DIFF
--- a/package.json
+++ b/package.json
@@ -655,6 +655,11 @@
                                 "pattern": "^(25[0-5]|2[0-4][0-9]|1[0-9][0-9]|[1-9]?[0-9])\\.(25[0-5]|2[0-4][0-9]|1[0-9][0-9]|[1-9]?[0-9])\\.(25[0-5]|2[0-4][0-9]|1[0-9][0-9]|[1-9]?[0-9])\\.(25[0-5]|2[0-4][0-9]|1[0-9][0-9]|[1-9]?[0-9])$",
                                 "type": "string"
                             },
+                            "remoteIPAddress": {
+                                "default": null,
+                                "description": "IP Address for remote target selection in gdb",
+                                "type": "string"
+                            },
                             "serialNumber": {
                                 "default": null,
                                 "description": "J-Link Serial Number - only needed if multiple J-Links are connected to the computer",
@@ -1225,6 +1230,11 @@
                                 "default": null,
                                 "description": "IP Address for networked J-Link Adapter",
                                 "pattern": "^(25[0-5]|2[0-4][0-9]|1[0-9][0-9]|[1-9]?[0-9])\\.(25[0-5]|2[0-4][0-9]|1[0-9][0-9]|[1-9]?[0-9])\\.(25[0-5]|2[0-4][0-9]|1[0-9][0-9]|[1-9]?[0-9])\\.(25[0-5]|2[0-4][0-9]|1[0-9][0-9]|[1-9]?[0-9])$",
+                                "type": "string"
+                            },
+                            "remoteIPAddress": {
+                                "default": null,
+                                "description": "IP Address for remote target selection in gdb",
                                 "type": "string"
                             },
                             "serialNumber": {
--- a/src/common.ts
+++ b/src/common.ts
@@ -129,7 +129,8 @@ export interface ConfigurationArguments extends DebugProtocol.LaunchRequestArgum
     ipAddress: string;
     serialNumber: string;
     jlinkscript: string;
-    
+    remoteIPAddress: string;
+
     // OpenOCD Specific
     configFiles: string[];
     searchDir: string[];
@@ -138,7 +139,7 @@ export interface ConfigurationArguments extends DebugProtocol.LaunchRequestArgum
 
     // PyOCD Specific
     boardId: string;
-    
+
     // StUtil Specific
     v1: boolean;
 
--- a/src/jlink.ts
+++ b/src/jlink.ts
@@ -28,12 +28,13 @@ export class JLinkServerController extends EventEmitter implements GDBServerCont
     public customRequest(command: string, response: DebugProtocol.Response, args: any): boolean {
         return false;
     }
-    
+
     public initCommands(): string[] {
         const gdbport = this.ports[createPortName(this.args.targetProcessor)];
+        const remoteip = (this.args.remoteIPAddress) ? this.args.remoteIPAddress : "localhost";
 
         return [
-            `target-select extended-remote localhost:${gdbport}`
+            `target-select extended-remote ${remoteip}:${gdbport}`
         ];
     }
 
@@ -77,7 +78,7 @@ export class JLinkServerController extends EventEmitter implements GDBServerCont
         const portMask = '0x' + calculatePortMask(this.args.swoConfig.decoders).toString(16);
         const swoFrequency = this.args.swoConfig.swoFrequency | 0;
         const cpuFrequency = this.args.swoConfig.cpuFrequency | 0;
-        
+
         const commands: string[] = [
             `monitor SWO EnableTarget ${cpuFrequency} ${swoFrequency} ${portMask} 0`,
             'DisableITMPorts 0xFFFFFFFF',
@@ -87,7 +88,7 @@ export class JLinkServerController extends EventEmitter implements GDBServerCont
         ];
 
         commands.push(this.args.swoConfig.profile ? 'EnablePCSample' : 'DisablePCSample');
-        
+
         return commands.map((c) => `interpreter-exec console "${c}"`);
     }
 
@@ -105,7 +106,7 @@ export class JLinkServerController extends EventEmitter implements GDBServerCont
             }
         }
     }
-    
+
     public serverArguments(): string[] {
         const gdbport = this.ports['gdbPort'];
         const swoport = this.ports['swoPort'];


haneefdm:
May I ask how this is different from using the `external` type gdb-server. It allows the server to be anywhere in the universe and use whatever server. It is the user's responsibility to start and manage the server and what IP address to use.

We have a lot of people using WSL2 already and it has its own issues. All we need is an IP address with the `external` server.

tamisoft:
This is different, because the jlinkgdbserver process gets a properly managed lifecycle. I am using vscode+jlink+wsl2with 2-3 different processors, eg. NRF52840, STM32F0-F4s and in many occasions the gdbserver locks up, and since it has no option to kill the connections forcefully, they start piling up causing weird issues. (I mean in the GDB line it'll show stuck tcp connections, probably the gdb clients are stuck after pressing the STOP in the debugging session). I suspect when the processors get exceptions, and the exception handler has a reboot, but since some peripherals might not reset properly the next startup will trigger again, and so goes an endless boot loop overloading the parser.
So this helps by starting always fresh, and also I am not sure what's the point ever to hardcode such things as a hostname :)

So far this way it works as intended where the compiler toolchain and gdb client comes from wsl2's linux side, and the jlinkgdbservercl.exe comes from the host's jlink installation.

haneefdm:
What I mean is that `external` server works with another setting called `gdbTarget` which sounds very similar to your `remoteIPAddress`.  The difference is you have to specify both `host:port`. This is precisely why we created the `external` server mode to handle all the conditions we cannot predict

I don't see how your code can even work because you are letting the extension find free ports on the localhost but then changing the hostname/IP. We go through an elaborate process to find free ports on the localhost. I must be missing something. We don't currently have the ability to do the same with remote hosts

JLink also has a single run mode where you can enable/disable it.

image

https://wiki.segger.com/J-Link_GDB_Server

Normally, it is okay to use the JLink server in the multi-run mode

tamisoft:
I think the external mode doesn't spawn the debugger, right? In that case it doesn't matter if it is single run or not, because it has to be managed manually on the host OS. And the sweetness of type=jlink is gone. (pressing F5..debug..shift F5.. code, rinse and repeat)

I checked the code for the port finding, you are right about that, probably that is another option that could be added to override the automatic behavior and specify the ports. It is not covering the other ports the jlinkgdbserver opens: 19021 19030, 19080 etc. I am actively using the RTTTelnetport.

So the key is that jlink as a type has the awesomeness of spawning and killing the debugger, external doesn't.
What do you think? What would be the best next step?

haneefdm:
The external mode does spawn the debugger (gdb). It does not spawn the gdb-server (OpenOCD, JLInk, or whatever).

I have to think about the rest

tamisoft:
Sorry of course I meant the gdb server. The client part is always fine as far as I can tell.

haneefdm:
I am going to reject this PR. There are already ways to do this properly with the "external" type gdb server and "gdbTarget" to specify the ip-address/port

tamisoft:
Can you please give me then the equivalent configuration if it is already there?

haneefdm:
See the Changelog first bullet (other bullets help as well)

https://github.com/Marus/cortex-debug/blob/master/CHANGELOG.md#v027

In launc.json, It looks something like
```JSON
    "servertype": "external",
    "gdbTarget": "host:port-num"
```
the host can be anything/anywhere ip-address, hostname so long as your TCP/IP can resolve it.

tamisoft:
I understand, but as you can see above it doesn't fit the case, just doesn't cut it. I wouldn't have had created this change if it was possible to make it work with the already existing parameters. If you don't want to introduce a new parameter, why not allow the gdbTarget for jlink type - removing the hard coded "localhost".

haneefdm:
I still don't get it. You are not giving me enough info on what is not working. I don't buy your argument about the lifecycle. If there is a bug with JLink, you should file it there. There are many people using this extension with WSL2

https://github.com/Marus/cortex-debug/issues/308

Marus:
First - sorry about not looking at this sooner - just haven't had time to dedicate to the extension in a long long time. 

I think I understand what @tamisoft is saying about this issue in regards to WSL2 - if I can summarize I believe it is the following:

-  has the toolchain environment setup inside a WSL2 environment (I am a little bit fuzzy on why you'd do this - I can't really think of a major advantage but I'm sure there are some advantages - if for nothing else a bit easier access to scripting environments like bash and stuff)
- WSL2 doesn't allow access to some (maybe all) USB devices - so can't run the native J-Link server within WSL2 - can only access the debug probe via the windows layer J-Link GDB server
- Could use the external type - but this has disadvantages of cortex-debug not automatically launching the GDB server with appropriate parameters, also wouldn't be able to use things like the direct SWO output and stuff
- The extension can launch the windows J-Link server - but can't access the server as with WSL2 "localhost" within the environment is isolated from the windows localhost.

@tamisoft - can you give me a bit more detail of exactly how you went about setting up your environment - I'm not opposed to some configuration options to this - but I kind of like the idea of a WSL2 flag vs the current approach taken in this PR.

RisinT96:
As you said yourself, you can launch the gdbserver from within linux, but you can't connect to it on localhost, correct.

But you can connect to it via the host's ip which can be found in /etc/resolv.conf.

Just have to add the gdbserver to windows' firewall.

Alternatively, what works for me is launching the jlink remote server on windows and leaving it open in the background, and adding the `ip` parameter to the `launch.json` with the host's ip (actually an environment variable that's created from /etc/resolve.conf in my .bashrc)

TorBlackbeard:
With Segger/JLink tools there's a third way:

Start "J-Link Remote Server" on the windows side.
Edit launch.json to have this:
            "servertype": "jlink",
            "serverpath": "JLinkGDBServerCLExe",
            "ipAddress": "192.168.1.10", <-------- substitute with your IP address

This requires you to be able to "ping 192.168.1.10" from a WSL terminal. This can be blocked by McAfee firewall. If you have that problem, and can not disable McAfee, you're forced to examine all kinds of advices on how to bridge the networks between WSL and windows in the HyperV". But all this is no different than accessing a (JLink)gdbServer by IP on the windows side.

Also I have a VcXsrv (X-server) running on the windows side. This makes JLinkGDBServerCLExe actually paint the download-progress-bar. I have no idea, though, how it makes sense, that the CommandLine edition of the tool decides to paint a GUI...

TorBlackbeard:
> (WSL:) I am a little bit fuzzy on why you'd do this 
The reasons you mention yourself , plus;
- using the SAME tools on build-server as on client (test what you fly, fly what you test)
- WSL2 is just a remoting-target in the eyes of vscode. This works equally well in docker-containers, so in that case we can achieve container-based-workflows and use the .devcontainer file in vscode to automatically fetch, run, remote-into the correct container for this particular project.
- This in turn lets us get rid of host-tools on windows all-together. And ease working on different projects, and using effort on mangling with the environment to satisfy a build-pipeline that MUST have ruby 2.4 in path, and another might require 2.7 in path.
- Some companies have funny IT permissions, e.g. missing ability to install/update software in windows, but apt install in WSL works fine. Also performance is sometimes magnitudes faster in WSL... in case of strange IT policies, but such exist, believe me!

> so can't run the native J-Link server within WSL2 - can only access the debug probe via the windows layer J-Link GDB server
See the 3rd way above.

DDevine:
I was looking for an answer to this problem as I have this setup too. I'd like to see documentation and support improve around this use-case - I'd do it myself but I am still trying to understand things.

We are currently (mainly) using a CLion based development approach but it uses SFTP file syncing between a remote docker build host (which is slow and error prone), and the transparency of the configuration of the development environment is extremely poor so the reason things sometimes break can only be described as Magic. Changing between products and variants comes down to copying and pasting configuration.

My reason for using VS Code Remote Development is to  fix 3 main issues: 
1) Reliably reproducing the build chain on Windows is a pain. We have to support hardware for years and during this time compilers, dependencies and Windows changes which causes havoc.
2) We want to develop on the same setup as our build/test-servers. 
3) We need consistent environments between team members to help with sharing tasks and reducing the "bus-factor" and to reduce time onboarding new members.
4) We have multiple products and multiple variants of those products all need slight configuration changes - so using VS code workspaces simplifies beginning work on any given product down to just selecting the desired workspace.
5) Many members would like to get away from Windows as it is a constant pain in the ass - so making sure our day-to-day work is not affected by containerising the workflow is a great way to do this... The only hang-up is the debugger. On Linux hosts this is an easy fix as you can just map /dev/ttyusb onto the contianer but, on Windows you need to run a GDB server (eg. JLinkServer) on the host but it can just be loaded on Start-up/log-in.

I'm going to see if @TorBlackbeard's configuration suggestion works - if so, that's ideal! 

By the way if you have both WSL2 and Docker installed on Windows, you can use "host.docker.internal" to reliably point to the host inside WSL2! New docker uses WSL2 as its virtualisation backend - so they are integrated.

wbober:
I'd like to back up this feature and the reasons posted in the topic for using WSL2. Most of them apply to our situation as well (particularly performance differences, easy access to available tools, etc.). While starting a debug session can be achieved using the external config or simply by manipulating iptables to forward traffic from WSL localhost to the external IP it is a bit cumbersome.

@Marus you have asked about the setup in WSL2. What I did is a simple script that executes the jlink process in windows. This is a WSL2 feature where you can start windows executables directly from the WSL. This way I can start the JLINK as it would be done natively but the process fails on connecting as the IP must be set to the WSL host.

```
#!/bin/bash

/mnt/c/Program\ Files\ \(x86\)/SEGGER/JLink/JLinkGDBServer.exe $*
```

How can I help to make this feature happen?

DanieleNardi:
I spent some time to [open an issue](https://github.com/Marus/cortex-debug/issues/451), then I've found this pull request. What's the status of this improvement?

haneefdm:
Going to close this. This is not the right way and while I have a better way, at least for WSL, even that may be obsolete. Several users are now also using `usbip`. I have seen JLink users use it inside WSL2 with the linux version of the gdb-server successfully. Back when this was submitted, `usbip` was still not working now, for some people it seems to work. MS is pitching in for `usbip` to work as well (rumors)

See #617

PR #342, opened 2020-10-08
Initial proof of concept for FreeRTOS task information

microwavesafe:
This is a first attempt at putting FreeRTOS aware debugging into the cortex-debug extension. I have never controlled gdb directly or written / modified a VSCode extension before, so I may have made some fundamental errors. I am hoping that the community that uses this extension is interested in these features and that we can shape what I have started into something useful for everyone.

First thing I need to work out is how to only turn on the FreeRTOS view if the rtos = 'FreeRTOS' option is set. At the moment I can't see how this is done as the view is set in package.json. Is there some way to hide / disable it?

I would also like to extend the functionality to include stack free space as this is invaluable. Performance data can also be included with the right config.

I am not expecting this to be merged as is, but I was hoping I can get some direction from the maintainers as to where this needs to go to be accepted.

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,6 +51,13 @@ Requirements:
 
 See https://github.com/Marus/cortex-debug/wiki for usage information. This needs some help from the community
 
+## Development
+
+Clone the Git repository, open the folder in VSCode and open a terminal. Run
+$ npm install
+$ npm run-script compile
+Now hit F5 to open a new VSCode window with the development extension running, you can load an existing project and launch the debugger for testing.
+
 ## Acknowledgments
 
 Parts of this extension are based upon Jan Jurzitza's (WebFreak) code-debug extension (https://github.com/WebFreak001/code-debug). His project provided an excellent base for GDB MI parsing and interaction.
--- a/package.json
+++ b/package.json
@@ -1505,6 +1505,11 @@
                     "id": "cortex-debug.registers",
                     "name": "Cortex Registers",
                     "when": "debugType == cortex-debug"
+                },
+                {
+                    "id": "cortex-debug.freertos",
+                    "name": "FreeRTOS Tasks",
+                    "when": "debugType == cortex-debug"
                 }
             ]
         }
--- a/src/frontend/extension.ts
+++ b/src/frontend/extension.ts
@@ -4,6 +4,7 @@ import * as path from 'path';
 
 import { PeripheralTreeProvider } from './views/peripheral';
 import { RegisterTreeProvider } from './views/registers';
+import { FreeRTOSTreeProvider } from './views/freertos';
 import { BaseNode, PeripheralBaseNode } from './views/nodes/basenode';
 
 import { SWOCore } from './swo/core';
@@ -35,9 +36,11 @@ export class CortexDebugExtension {
     private peripheralProvider: PeripheralTreeProvider;
     private registerProvider: RegisterTreeProvider;
     private memoryProvider: MemoryContentProvider;
+    private freeRTOSProvider: FreeRTOSTreeProvider;
 
     private peripheralTreeView: vscode.TreeView<PeripheralBaseNode>;
     private registerTreeView: vscode.TreeView<BaseNode>;
+    private freeRTOSTreeView: vscode.TreeView<BaseNode>;
 
     private SVDDirectory: SVDInfo[] = [];
     private functionSymbols: SymbolInformation[] = null;
@@ -46,6 +49,7 @@ export class CortexDebugExtension {
         this.peripheralProvider = new PeripheralTreeProvider();
         this.registerProvider = new RegisterTreeProvider();
         this.memoryProvider = new MemoryContentProvider();
+        this.freeRTOSProvider = new FreeRTOSTreeProvider();
 
         let tmp = [];
         try {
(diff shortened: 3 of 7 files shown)


Marus:
Sorry for leaving this hanging for such a long time. This looks like the start of something very good and useful - I'll try to get a chance to look at this in more detail and see if I think it is something we can pursue further.

microwavesafe:
No problem at all. I didn't want to go any further in case it was either unwanted or there was a better way of handling the requests. The GDB interface code is complicated, so I really wasn't sure if there was an easy way of access the required variables.

I was hoping to separate of the FreeRTOS code completely, but I couldn't get this to work at all, so ended up modifying the GDB code with a custom function.

Anyway let me know what you think the next steps are.

haneefdm:
Hmmm, I use Cortex-Debug with FreeRTOS and other rtos's all the time. I see all the threads and their frames. So, why is this change needed? There a dozen RTOS's. Could you explain the following?

`I would also like to extend the functionality to include stack free space as this is invaluable. Performance data can also be included with the right config.`

haneefdm:
The RTOS data structures seem to change very frequently. But the gdb/gdb-server remains constant

microwavesafe:
This is only a nicety. It is not necessary and all information can be gathered in run time using functions such as ```uxTaskGetSystemState```. There are times where knowing when a task has inherited a higher priority or is blocking is very useful. Having it always at hand when you need it is easier to use than writing specific debug code per application, especially if you have a hit a bug that only occurs very occasionally. 

Mainly I use free stack space for tuning the stack size of each task I create. Often my designs are heavily RAM constrained, so it is very useful to see which tasks can safely have their stack size reduced and which are up against their limits and could do with a bit more.

microwavesafe:
I can't comment on other RTOS, but the FreeRTOS task struct is mostly unchanged for a long time now.

haneefdm:
How about using the watch window to display task info? Does that not work?

Yes, FreeRTOS data structures/variables have changed. In fact it broke OpenOCD and other gdb servers.

microwavesafe:
You can use a watch window, I just find it easier to use to have the information on hand, rather than manually setting up a watch for each task. It also allows us to display the data in a more readable format.

Anything that can make my day to day developing experience better, is something worth exploring, in my opinion.

I'm happy to close this pull request if you don't think it fits. It's no problem.

DNedic:
I for one would really appretiate having a more advanced FreeRTOS debugging expirience, especially since it uses obfuscation for control blocks for instance and debugging with a watch window is a pain.

Marus:
@haneefdm does bring up a good point about maintenance- as it will break with changes to the structure- but changes to the the FreeRTOS task structure do seem to be fairly infrequent at least (although for certain not non-existing). And we do have to consider what version(s) of FreeRTOS we are going to support. One possibility is to only support the current LTS version - which should at least reduce the likelihood of task structure changes. 

As to getting it to only show for FreeRTOS debug sessions - I’ll have to think about that. Ideally we’d be able to set a context variable which the package.json can reference to determine visibility- but to be honest I’m not sure if we can add the appropriate context variables- going to have to dig a bit on that one. (If I can figure that out though it would be nice as would allow hiding the peripheral register view if the user doesn’t have an SVD definition).  

I’ll try to play with this a bit and see if I can think of any other improvements that would need to be made before including it - but certainly I think it is off to a good start - and do see value in this kind of summary info on top the regular thread/stack frames that would currently be displayed.

microwavesafe:
I was wondering about this the other day. Would it be possible to bring the required GDB functions out to some kind of API that we can then use to implement inspection add ons. If we can do this in a way that is easy, then implementing and maintaining the add ons will be a more achievable goal.

PhilippHaefele:
I would also suggest adding an API instead of adding specific implementations directly into Cortex Debug. Not all RTOS have things like functions for getting free task stack, so at some points there is a need for having such a RTOS plugin.
Segger Ozone has a javascript API for showing RTOS values in a flexible way. Maybe it‘s a starting point to look into this?
@haneefdm What do you think?

haneefdm:
@PhilippHaefele Yup, that is why this PR is sitting here. It needs to be more generic and perhaps even user-customizable or users can make their own.

microwavesafe:
I found the GDB interface code hard going. But a generic way of accessing memory by variable name would go a long way to making an API. It seems overly complicated to do this, unless I am missing something obvious?

microwavesafe:
Also there didn't seem to be much appetite for such a change at the time I create this, so I just left it.

microwavesafe:
I have settled on using a gdb script for this. Note this is not generic and will only work for FreeRTOS of specific versions, but it works perfectly for me so I thought I'd share. It is adapted from an existing script I found [.gdbinit-FreeRTOS-helpers](https://github.com/ErichStyger/McuOnEclipse_PEx/blob/master/Drivers/freeRTOS/gdbBacktraceDebug/.gdbinit-FreeRTOS-helpers)

File attached. I will close this pull request as it is unlikely to go anywhere and the script I've attached fulfils my needs
[gdbinit-FreeRTOS.zip](https://github.com/Marus/cortex-debug/files/8185253/gdbinit-FreeRTOS.zip)

PR #360, opened 2020-11-24
Toggle disabling and enabling FunctionBreakPoints.

kohada88:
Currently, FunctionBreakPoints cannot be disabled even if the check box is cleared.
This fix toggles between disabling and enabling FunctionBreakPoints with checkbox.


Marus:
The 0.3.8-pre release also includes this change as well.

PR #363, opened 2020-12-01
delete all break points with disconnect.

kohada88:
Some targets still alive break-points even if disconnected.
When hits the break-points without a debugger, there is no way to continue with the target still stopped.


RisinT96 (approved):

Marus:
Hi @kohada88  - I've merged this into master and have uploaded a 0.3.8-pre.visx to the release page on GitHub.

Should hopefully be doing a real 0.3.8 release in the next few days - just have a few things to test and verify before I do.

Issue #354, opened 2020-11-10
Official ST-LINK gdbserver support


resonantworks:
ST's ST-LINK gdbserver from STM32CubeIDE seems to be a lot more stable than OpenOCD for the devices I have been using.

[ST-LINK gdbserver manual](https://www.st.com/resource/en/user_manual/dm00613038-stm32cubeide-stlink-gdb-server-stmicroelectronics.pdf)

On Windows, I run the st-link gdbserver in a separate command line via:
```cmd
C:\Apps\stm32\STM32CubeIDE\plugins\com.st.stm32cube.ide.mcu.externaltools.stlink-gdb-server.win32_1.4.0.202007081208\tools\bin\ST-LINK_gdbserver.exe -p 3333 -l 1 -d -s -i 066CFF495652716587091651 -e -cp C:\Apps\stm32\STM32CubeIDE\plugins\com.st.stm32cube.ide.mcu.externaltools.cubeprogrammer.win32_1.4.0.202007081208\tools\bin -m 0 -k
```
Param|Option
-- | --
-l | logging verbosity
-d | SWD mode
-s | Flash verify
-i | ST-LINK serial number (optional - will choose the first found when left out)
-e | Persistent (do not quit - most likely do not want this when launching within vscode)
-cp | CubeProgrammer plugin bin path
-m | 'AppID to debug' (this was grabbed straight from the params STM32CubeIDE supplied launching st-link gdbserver)
-k | Initialise under reset

It also lists options for SWO (which I have yet to get around to trying):
Param|Option
-- | --
-z | SWO TCP port
-a | MCU clock speed
-b | Clock divider

I use something along the following in the launch.json settings:
```json
{
    "name": "stlink-gdbserver F401RE",
    "type": "cortex-debug",
    "request": "launch",
    "runToMain": true,
    "cwd": "${workspaceRoot}",
    "executable": "${command:cmake.launchTargetPath}",
    "servertype": "external",
    "gdbTarget": ":3333",
    "device": "STM32F401RETx",
    "svdFile": "${workspaceRoot}/Environment/svd/STM32F401.svd",
}
```

If Cortex-Debug had a new config setting for the st-link gdbserver path, then I'm sure the rest could be pattern matched from that. So given:  
`C:\Apps\stm32\STM32CubeIDE\plugins\com.st.stm32cube.ide.mcu.externaltools.stlink-gdb-server.win32_1.4.0.202007081208\tools\bin\ST-LINK_gdbserver.exe`

[...]

PR #365, opened 2020-12-02
Added initial support for STMicroelectronic's ST-LINK gdbserver

hwmaier:
After commenting last night that it would be nice to have that feature, I gave it a go to implement it. The only issue is that ST's STMCubeIde needs to be installed as a dependency and ST does not make it easy to determine the correct path names for the 3 tools involved (namely `ST-LINK_gdbserver`, `arm-none-eabi-gdb`, `STM32_Programmer`) . So those have to be added by hand by the user as settings. I wish this could be done automatically.

So either the tools are on the PATH or they have to be configured like this:
```
           "armToolchainPath": "c:/ST/STM32CubeIDE_1.5.0/STM32CubeIDE/plugins/com.st.stm32cube.ide.mcu.externaltools.gnu-tools-for-stm32.7-2018-q2-update.win32_1.5.0.202011040924/tools/bin",
            "stm32cubeprogrammer": "c:/ST/STM32CubeIDE_1.5.0/STM32CubeIDE/plugins/com.st.stm32cube.ide.mcu.externaltools.cubeprogrammer.win32_1.5.0.202011040924/tools/bin",
            "serverpath": "c:/ST/STM32CubeIDE_1.5.0/STM32CubeIDE/plugins/com.st.stm32cube.ide.mcu.externaltools.stlink-gdb-server.win32_1.5.0.202011040924/tools/bin/ST-LINK_gdbserver.exe"
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,7 +45,7 @@ Requirements:
     * OpenOCD - provides a GDB Server that can be used with a number of debuggers (http://openocd.org)
         * NOTE: On macOS do not use the default version of OpenOCD provided by homebrew, this is not compatible with releases V0.2.4 and newer. You can either install from source using homebrew (`brew install open-ocd --HEAD`) or the packages from https://github.com/gnu-mcu-eclipse/openocd/releases will also work. Some linux versions and Windows may also need a more up-to-date version of OpenOCD from the gnu-mcu-eclipse releases.
     * Texane's st-util GDB server - Only supports ST-Link Debug Probes (https://github.com/texane/stlink)
-    * ST-LINK GDB server - This server is packaged with the [STM32CubeIDE](https://www.st.com/en/development-tools/stm32cubeide.html). In addition the [STM32CubeProgrammer](https://www.st.com/en/development-tools/stm32cubeprog.html) tool must be installed.
+    * ST-LINK GDB server - This server is packaged with the [STM32CubeIDE](https://www.st.com/en/development-tools/stm32cubeide.html) which must be installed. The location of the STM32CubeIDE and related tools is automatically resolved but also can be overridden using configuration settings (`armToolchainPath`, `stm32cubeprogrammer` and `serverpath`).
     * pyOCD GDB Server - GDB server that supports the CMSIS-DAP debugger on a number of mbed boards (https://github.com/mbedmicro/pyOCD)
     * Black Magic Probe
 
--- a/README.md
+++ b/README.md
@@ -5,6 +5,7 @@
 Debugging support for ARM Cortex-M Microcontrollers with the following features:
 
 * Support J-Link, OpenOCD GDB Server
+* Initial support for STMicroelectronic's ST-LINK GDB server (no SWO support yet)
 * Partial support for PyOCD and textane/stlink (st-util) GDB Servers (SWO can only be captured via a serial port)
 * Initial support for the Black Magic Probe (This has not been as heavily tested; SWO can only be captured via a serial port)
 * Cortex Core Register Viewer
@@ -44,6 +45,7 @@ Requirements:
     * OpenOCD - provides a GDB Server that can be used with a number of debuggers (http://openocd.org)
         * NOTE: On macOS do not use the default version of OpenOCD provided by homebrew, this is not compatible with releases V0.2.4 and newer. You can either install from source using homebrew (`brew install open-ocd --HEAD`) or the packages from https://github.com/gnu-mcu-eclipse/openocd/releases will also work. Some linux versions and Windows may also need a more up-to-date version of OpenOCD from the gnu-mcu-eclipse releases.
     * Texane's st-util GDB server - Only supports ST-Link Debug Probes (https://github.com/texane/stlink)
+    * ST-LINK GDB server - This server is packaged with the [STM32CubeIDE](https://www.st.com/en/development-tools/stm32cubeide.html). In addition the [STM32CubeProgrammer](https://www.st.com/en/development-tools/stm32cubeprog.html) tool must be installed.
     * pyOCD GDB Server - GDB server that supports the CMSIS-DAP debugger on a number of mbed boards (https://github.com/mbedmicro/pyOCD)
     * Black Magic Probe
 
--- a/package.json
+++ b/package.json
@@ -1294,7 +1294,7 @@
                             },
                             "stm32cubeprogrammer": {
                                 "default": null,
-                                "description": "Path to the STM32CubeProgrammer tool's bin directory. Either installed as part os STM32CubeIDE or separatly.",
+                                "description": "This path is normally resolved to the installed STM32CubeIDE or STM32CubeProgrammer but can be overridden here.",
                                 "type": "string"
                             },
                             "targetId": {
@@ -1375,10 +1375,7 @@
                             "name": "${6:Debug Microcontroller}",
                             "request": "launch",
                             "type": "cortex-debug",
-                            "servertype": "stlink",
-                            "armToolchainPath": "c:/ST/STM32CubeIDE_1.5.0/STM32CubeIDE/plugins/com.st.stm32cube.ide.mcu.externaltools.gnu-tools-for-stm32.7-2018-q2-update.win32_1.5.0.202011040924/tools/bin",
-                            "stm32cubeprogrammer": "c:/ST/STM32CubeIDE_1.5.0/STM32CubeIDE/plugins/com.st.stm32cube.ide.mcu.externaltools.cubeprogrammer.win32_1.5.0.202011040924/tools/bin",
-                            "serverpath": "c:/ST/STM32CubeIDE_1.5.0/STM32CubeIDE/plugins/com.st.stm32cube.ide.mcu.externaltools.stlink-gdb-server.win32_1.5.0.202011040924/tools/bin/ST-LINK_gdbserver.exe"
+                            "servertype": "stlink"
                         },
                         "description": "Debugs an embedded ARM Cortex-M microcontroller using GDB + STMicroelectronic's ST-LINK_gdbserver part of STM32CubeIDE",
                         "label": "Cortex Debug: ST-LINK"
--- a/package.json
+++ b/package.json
@@ -66,6 +66,14 @@
                     "default": null,
                     "description": "Path to the Texane's ST-Util GDB Server executable. If not set then st-util (st-util.exe on Windows) must be on the system path."
                 },
+                "cortex-debug.stlinkPath": {
+                    "type": [
+                        "string",
+                        "null"
+                    ],
+                    "default": null,
+                    "description": "Path to the ST-LINK_gdbserver executable. If not set then ST-LINK_gdbserver (ST-LINK_gdbserver.exe on Windows) must be on the system path."
+                },
                 "cortex-debug.enableTelemetry": {
                     "type": "boolean",
                     "default": true,
@@ -171,12 +179,13 @@
                         "properties": {
                             "servertype": {
                                 "type": "string",
-                                "description": "GDB Server type - supported types are jlink, openocd, pyocd, pe and stutil",
+                                "description": "GDB Server type - supported types are jlink, openocd, pyocd, pe, stlink, stutil, qemu and external",
                                 "enum": [
                                     "jlink",
                                     "openocd",
                                     "pyocd",
                                     "stutil",
+                                    "stlink",
                                     "bmp",
                                     "pe",
                                     "qemu",
@@ -657,12 +666,12 @@
                             },
                             "serialNumber": {
                                 "default": null,
-                                "description": "J-Link Serial Number - only needed if multiple J-Links are connected to the computer",
+                                "description": "J-Link or ST-LINK Serial Number - only needed if multiple J-Links/ST-LINKs are connected to the computer",
                                 "type": "string"
                             },
                             "interface": {
                                 "default": "swd",
-                                "description": "Debug Interface type to use for connections (defaults to SWD) - Used for J-Link and BMP probes.",
+                                "description": "Debug Interface type to use for connections (defaults to SWD) - Used for J-Link, ST-LINK and BMP probes.",
                                 "type": "string",
                                 "enum": [
                                     "swd",
@@ -742,12 +751,13 @@
                         "properties": {
                             "servertype": {
                                 "type": "string",
-                                "description": "GDB Server type - supported types are jlink, openocd, pyocd, pe and stutil",
+                                "description": "GDB Server type - supported types are jlink, openocd, pyocd, pe, stlink, stutil, qemu and external",
                                 "enum": [
                                     "jlink",
                                     "openocd",
                                     "pyocd",
                                     "stutil",
+                                    "stlink",
                                     "bmp",
                                     "pe",
                                     "qemu",
@@ -1229,12 +1239,12 @@
                             },
                             "serialNumber": {
                                 "default": null,
-                                "description": "J-Link Serial Number - only needed if multiple J-Links are connected to the computer",
+                                "description": "J-Link or ST-LINK Serial Number - only needed if multiple J-Links/ST-LINKs are connected to the computer",
                                 "type": "string"
                             },
                             "interface": {
                                 "default": "swd",
-                                "description": "Debug Interface type to use for connections (defaults to SWD) - Used for J-Link and BMP probes.",
+                                "description": "Debug Interface type to use for connections (defaults to SWD) - Used for J-Link, ST-LINK and BMP probes.",
                                 "type": "string",
                                 "enum": [
                                     "swd",
@@ -1279,8 +1289,13 @@
                             },
                             "v1": {
                                 "default": false,
-                                "description": "Set this to true if your debug probe is a ST-Link V1 (for example, the ST-Link on the STM32 VL Discovery is a V1 device). When set to false a ST-Link V2 device is used.",
+                                "description": "For st-util only. Set this to true if your debug probe is a ST-Link V1 (for example, the ST-Link on the STM32 VL Discovery is a V1 device). When set to false a ST-Link V2 device is used.",
                                 "type": "boolean"
+                            },
+                            "stm32cubeprogrammer": {
+                                "default": null,
+                                "description": "Path to the STM32CubeProgrammer tool's bin directory. Either installed as part os STM32CubeIDE or separatly.",
+                                "type": "string"
                             },
                             "targetId": {
                                 "description": "On BMP this is the ID number that should be passed to the attach command (defaults to 1); for PyOCD this is the target identifier (only needed for custom hardware)",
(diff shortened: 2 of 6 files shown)


matthew-brown-bluefruit:
Thanks for the initial effort! Could you not just use a single setting like the existing `cortex-debug.JLinkGDBServerPath`, and then use regex to build the CubeProgrammer path, ie:
```json
"cortex-debug.STLinkGDBServerPath": "${env:STLINK_GDBSERVER}",
```

See: https://github.com/Marus/cortex-debug/issues/354#issue-740119644

hwmaier:
Have you worked with the recent ST tools? There is no path relationship between the 3 components ST-LINK_gdbserver, arm-none-eabi-gdb, STM32_Programmer. ST organised them into different unrelated path structures with horrific longs names and embedded dated and version numbers.

No special configuration settings are required if the user has the ST-LINK_gdbserver.exe on the path and the STM32_Programmer installed into the default location. This is the easiest use case and the one I personally use.

The settings I mention above are only required if ST-LINK_gdbserver.exe is not on the path and/or the STM32_Programmer from the STM32CubeIDE are used rather the standalone version.

If the STM32CubeIDE is used then it is not possible to derive the settings from its location as the paths and sub-folders differ from version to version.

matthew-brown-bluefruit:
Hi, that is strange. I'm using a PowerShell script started by a vscode task which uses regex to work back from the st-link gdbserver path. I can't share the entire script as it's used in a work project, but here are the relevant bits
```powershell
...
$GdbServerRegx = "^(?<plugins>.+plugins).+stlink-gdb-server.(?<version>.+)\\t.+$"
...
function RunStLinkGdbServer($gdbServerPath) {
    PrintHeading "Running ST-LINK GdbServer .."

    $parsed = $gdbServerPath | Select-String -Pattern $GdbServerRegx
    if (!$parsed.Matches.Success) {
        Write-Error "Failed to parse ST-LINK GdbServer path"
        exit 4
    }

    $parts = $parsed.Matches[0]
    $pluginPath = $parts.Groups["plugins"].Value
    $version = $parts.Groups["version"].Value

    $programmerPath = "$pluginPath\com.st.stm32cube.ide.mcu.externaltools.cubeprogrammer.$version\tools\bin"

    $swoArgs = if ($script:swoFreq -gt 0) {
        $divider = $script:cpuFreq / $script:swoFreq
        @("-z", "$script:swoPort", "-a", "$script:cpuFreq", "-b", "$divider") 
    }
    else { @() }

    # Command line arguments: https://www.st.com/resource/en/user_manual/dm00613038-stm32cubeide-stlink-gdb-server-stmicroelectronics.pdf
    $connectFreqArgs = if ($script:connectFreq -gt 0) { @("--frequency", $script:connectFreq) } else { @() }
    Write-Host "$gdbServerPath -l 4 -d -s -m 0 -k -e -p $tcpPort -i $script:serial" @connectFreqArgs @swoArgs "-cp $programmerPath"
    & $gdbServerPath -l 4 -d -s -m 0 -k -e -p $tcpPort -i $script:serial @connectFreqArgs @swoArgs -cp $programmerPath
}
```
```

**Comment by hwmaier** [comment]
I see now what you mean. One indeed could use your reg-ex parsing concept to accomplish automatic detection of STM32CubeIDE. Thanks for sharing and if I find some time later I'll add it. The only thing which I am bit concerned is that it is closely tied to ST's current naming convention. If they change that in 6 month time, then it breaks everything. So some form of override should be available and I probably keep the current `stm32cubeprogrammer` setting to override the programmer location. The other two settings `armToolchainPath and `serverpath` are already existing settings, I didn't add those.

**Comment by matthew-brown-bluefruit** [comment]
Sounds like a good idea ;). I don't know vscode internals/typescript but the least I can do is find anything odd or typos in the PR. How can I go about testing this modification to the extension?

**Comment by hwmaier** [comment]
@matthew-brown-bluefruit I added auto-discovery of the STM32CubeIDE tools.

You can test the PR by pulling it into your ~/.vscode/extensions folder (and replacing the original cortex-debug) and compile using this command:

    npm run vscode:prepublish

When using the estension no special configuration settings required any more if STM32CubeIDE is installed. I tested this only on Windows, hopefully it works on Linux and Mac as well.

```
        {
            "name": "ST-Link Remote",
            "cwd": "${workspaceFolder}",
            "request": "launch",
            "type": "cortex-debug",
            "servertype": "stlink",
        },
```

**Comment by matthew-brown-bluefruit** [comment]
Excellent stuff! 😉 I'll give it a go on Windows and Linux and let you know.

**Comment by matthew-brown-bluefruit** [comment]
Hi, I've tested it successfully on Windows after a small fix. I've made a PR against your branch with the fix and some slight config improvements from my non-work github account: [Small stlink gdb server fix and tweaks](https://github.com/hwmaier/cortex-debug/pull/1).

This is the **launch.config** used to check:
```json
{
    "version": "0.2.0",
    "configurations": [
        {
            "name": "ST-Link Remote",
            "cwd": "${workspaceFolder}",
            "request": "launch",
            "type": "cortex-debug",
            "servertype": "stlink",
            "armToolchainPath": "C:\\Apps\\stm32\\STM32CubeIDE\\plugins\\com.st.stm32cube.ide.mcu.externaltools.gnu-tools-for-stm32.9-2020-q2-update.win32_1.5.0.202011040924\\tools\\bin",
            "stlinkPath": "C:\\Apps\\stm32\\STM32CubeIDE\\plugins\\com.st.stm32cube.ide.mcu.externaltools.stlink-gdb-server.win32_1.5.0.202011040924\\tools\\bin\\ST-LINK_gdbserver.exe",
            "stm32cubeprogrammer": "C:\\Apps\\stm32\\STM32CubeIDE\\plugins\\com.st.stm32cube.ide.mcu.externaltools.cubeprogrammer.win32_1.5.0.202011040924\\tools\\bin"        
        }
    ]
}
```


I'll see if can test it on Linux too...

matthew-brown-bluefruit:
On Ubuntu 20.04 desktop, it fails to launch the stlink gdbserver. I've installed STM32CubeIDE using the 'Generic Linux Installer' via sudo to `~/apps/stm32cudeide`, and then verified it runs correctly.

**Adapter Output**
```
/home/matthew/apps/stm32cudeide/plugins/com.st.stm32cube.ide.mcu.externaltools.stlink-gdb-server.linux64_1.5.0.202011040924/tools/bin/ST-LINK_gdbserver: error while loading shared libraries: libSTLinkUSBDriver.so: cannot open shared object file: No such file or directory
```

I get the same error from a terminal when trying to run the stlink gdbserver outside of its folder. That library is located here:
```
/home/matthew/apps/stm32cudeide/plugins/com.st.stm32cube.ide.mcu.externaltools.cubeprogrammer.linux64_1.5.0.202011040924/tools/lib/libSTLinkUSBDriver.so
/home/matthew/apps/stm32cudeide/plugins/com.st.stm32cube.ide.mcu.externaltools.stlink-gdb-server.linux64_1.5.0.202011040924/tools/bin/native/linux_x64/libSTLinkUSBDriver.so
```

**launch.json**
```json
{
    "version": "0.2.0",
    "configurations": [
        {
            "name": "ST-Link Remote",
            "cwd": "${workspaceFolder}",
            "request": "launch",
            "type": "cortex-debug",
            "servertype": "stlink",
            "armToolchainPath": "/home/matthew/apps/stm32cudeide/plugins/com.st.stm32cube.ide.mcu.externaltools.gnu-tools-for-stm32.7-2018-q2-update.linux64_1.5.0.202011040924/tools/bin",
            "stlinkPath": "/home/matthew/apps/stm32cudeide/plugins/com.st.stm32cube.ide.mcu.externaltools.stlink-gdb-server.linux64_1.5.0.202011040924/tools/bin/ST-LINK_gdbserver",
            "stm32cubeprogrammer": "/home/matthew/apps/stm32cudeide/plugins/com.st.stm32cube.ide.mcu.externaltools.cubeprogrammer.linux64_1.5.0.202011040924/tools/bin",
        }
    ]
}
```

I don't think this is easily solvable unless you can change the working directory to where the stlink gdbserver is before running it.

hwmaier:
@matthew-brown-bluefruit The "automatic" discovery of the location of the STMCubeIDE works only if it is installed in the default locations. On Windows this would be `c:\ST` and on Linux/Mac `/opt/st`. Could you test that too? I saw to you installed STMCubeIDE  into non-default locations. Then the 3 settings `armToolchainPath`, `stlinkPath` and `stm32cubeprogrammer` would not be needed.

The shared library issue on Linux is another problem for another day and I am not sure how it is solved for other gdbservers as this would not be unique to the ST-Link server.

hwmaier:
@matthew-brown-bluefruit: The gdbserver is started by this code: `this.server = new GDBServer(this.args.cwd, executable, args, initMatch, gdbPort);` which means one can configure the cwd.  Could you try if it launches if the `cwd` setting is set to the location of the ST-Link_gdbserver executable rather the `${workspaceFolder}`? This may however have other side-effects on gdb being able to resolve source file names and may not be a proper solution, but the information can help to find a work-around.

I can think of these options to solve the launch on Linux:

- Passing the additional environment variable LD_LIBRARY_PATH upon launch. Problem is that the location of libSTLinkUSBDriver.so is platform dependant. We could add all possible platforms to the LD_LIBRAY_PATH. Like `.../native/linux_x64:.../native/linux_x86` etc. I guess there is only Linux and Mac and x86, arm, and x64 as processors. But this is still a maintenance problem when platform names change.
- Adding the location of libSTLinkUSBDriver.so to the /etc/ld.so.conf.d/ files. Also complicated and we would have to change system files.
- Approach ST Microelectronic to fix the issue. Not sure if they would listen...
- Change the `GDBServer` class in `server.ts` to use as CWD not the one configured in the launch settings (use that for arm-none-eabi-gdb.exe) but use the `serverpath` as CWD when launching. This however may have unintended consequences for other gdbservers.

Example change from:
```
this.process = ChildProcess.spawn(this.application, this.args, { cwd: this.cwd });
```
to something like this (untested due to lack of a Linux system)
```
const env = process.env;
env.LD_LIBRAY_PATH = path.dirname(this.application)+'/native/linux_x64' + ':' + path.dirname(this.application)+'/native/linux_amd64'; // TODO: Add Mac variants as well
this.process = ChildProcess.spawn(this.application, this.args, { cwd: this.cwd, env: env });
```
or
```
this.process = ChildProcess.spawn(this.application, this.args, { cwd: path.dirname(this.application) || this.cwd });
```

matthew-brown-bluefruit:
@hwmaier I've checked that the autodiscovery works when installed using the installers default C:\ST folder. I'll have to leave the Linux side for the moment as I can't seem to debug that file (backend). After running `npm run compile`, whenever I try use the `Debug Server` launch it just spits out this:
```
C:\Program Files\nodejs\node.exe --nolazy .\src\gdb.ts --server=4711
(node:21952) Warning: To load an ES module, set "type": "module" in the package.json or use the .mjs extension.
(Use `node --trace-warnings ...` to show where the warning was created)
internal/process/warning:43
Uncaught C:\prj\github\cortex-debug\src\gdb.ts:1
import { DebugSession, InitializedEvent, TerminatedEvent, ContinuedEvent, OutputEvent, Thread, ThreadEvent, StackFrame, Scope, Source, Handles, Event } from 'vscode-debugadapter';
^^^^^^

SyntaxError: Cannot use import statement outside a module
Process exited with code 1
```

I tried googling, a few tweaks but no joy. If you know how to debug the backend please let me know.

Marus:
Can you test this out with the 0.3.8-pre2.vsix file that is in the GitHub release. I've made some changes to hopefully add support for macOS (although I can't actually get the STLink server to run). But added the appropriate base paths for macOS (assuming standard installation locations) and also reworked the `resolveCubePath` function so that it would work on a node v8 runtime (which is what the debug adapter is using on my macOS install at least - need to confirm that on other platforms)

hwmaier:
@Marus Have you seen the discussion further up re shared libraries on Linux (https://github.com/Marus/cortex-debug/pull/365#issuecomment-743924301) ? My guess on OSX this is the same issue.

hwmaier:
@Marus 
> Can you test this out with the 0.3.8-pre2.vsix file that is in the GitHub release

Tested tonight using 0.3.8-pre2.vsix. Hit F5 and everything works. This is on Windows 10 using STM32CubeIDE_1.5.0 installed at default location. Very happy seeing this becoming part of the official release.

Marus:
Glad to hear it is working for you on Windows.

Yeah - I expect it is related to the same shared libraries issue that you had pointed out - just need to dig in it more to try and figure out if there is some way to work around that issue.

I think I will release it as it currently stands noting for users that currently this server is only supported on a Windows host environment; and hopefully at some point myself (or someone else) will be able to determine why the shared libraries for this server don't want to load from the VS code environment.

yueduz:
Failed to launch ST-LINK GDB Server: Error: spawn /opt/stm32cubeide/plugins/- com.st.stm32cube.ide.mcu.externaltools.stlink-gdb- server.linux64_15.0.202011040924/tools/bin EACCES
2021-01-23 10-13-56 的屏幕截图
2021-01-23 10-15-38 的屏幕截图

hwmaier:
@yueduz There are known issue with Linux and MaxOSX relating to shared libs, refer to explanation further up in this thread https://github.com/Marus/cortex-debug/pull/365#issuecomment-743942717 and hence the support is currently limited to Windows until a solution is found. I am not sure if your report relates to that (eg shared lib cannot be loaded) or there is another issue with file privileges, the EACCESS error hints that. Can you check if you have exec rights on the plugin?

yueduz:
```
export LD_LIBRARY_PATH=/opt/stm32cubeide/plugins/com.st.stm32cube.ide.mcu.externaltools.cubeprogrammer.linux64_1.5.0.202011040924/tools/lib:$LD_LIBRARY_PATH
```

It can be used to temporarily add library search path in linux

lgrigoryeva1:
So did I understand right that there's no support at all on Linux at the moment? @hwmaier

hwmaier:
@lgrigoryeva1 If you manage to get the shared library path configured as mentioned above, it should work. The issue is that we have no means currently to automatically set the shared library path before launch. So to answer your question, no official out-of-the-box support on Linux yet, but experimental support for developers who understand how to configure shared library paths on Linux.

yueduz:
https://www.st.com/content/st_com/zh/products/development-tools/software-development-tools/stm32-software-development-tools/stm32-performance-and-debuggers/st-link-server.html

At present, the official has provided a separate installation package, which supports out-of-the-box use

hwmaier:
@yueduz  This is an ST-Link server and has been around for a while. Its purpose is to  share the ST link device amongst different apps. It is not a GDB server, you would still have to install a configure a GDB server.

yueduz:
2021-12-18 19-12-58 的屏幕截图
Hi!
I found that the path of the shared library can be set for linux, but how to provide it as a configuration parameter? I don't know anything about TS, who can do this work?

yueduz:
I want to use "console.log()" to print the log in "start", but I don't see the output in the terminal。

hwmaier:
@yueduz  This appears to be a feasible approach indeed. I would suggest to derive the shared lib path from the location of the executable to avoid the introduction of another configuration parameter. We already have too many for my opinion. Try using the directory name of `stlinkPath` for example.

yueduz:
````js
{
            "stm32cubeide":"/opt/stm32cubeide/ ,
            "stm32cubeprogrammer": "/opt/stm32cubeprog/"
}
```
Then I think it is better to use these two parameters, and it looks neater. Then cortex-debug finds the things needed in it.

Only configure to the top-level directory of the software installation.

haneefdm:
@yueduz Are you using VSCode to debug the Extension?. If you are, then look in the "Debug Console" of that Instance of VSCode for the output of your console.log().

I see syntax errors in the code above (no semi-colons at the end) and also some other linting issues (uses of spaces or lack thereof). TSLint should have caught these.

You are modifying the parent processes env. Please don't do that. Make a copy and use that. or you are messing with the adapter processes' env.

This code should also be linux specific.

yueduz:
I tried to hit a breakpoint, but why didn't it work? When I run it will not pause at the breakpoint
2021-12-19 09-17-41 的屏幕截图
.

haneefdm:
You missed a couple of steps...

https://github.com/Marus/cortex-debug#how-to-debug

You need to debug both parts of the extension

yueduz:
I think I have done the steps in the link, except for switching branches. Then I hit two breakpoints, one in the js file and one in the ts file. The breakpoint in ts has no effect. Does the ts file do not support breakpoints
2021-12-19 10-25-44 的屏幕截图
?

haneefdm:
image

mi2.ts is in another program and not the Extension. You need to debug both.
The above is important. Hopefully, you also modified the FW launch.json with the 4711 port.

yueduz:
I did not do this: json with the 4711 port.
Now I can debug normally.
My English level is worse than my ts level, and I have been using Google Translate to translate our conversations.
thank you very much for your help.

haneefdm:
No worries. You are doing fine.

hwmaier:
> ```js
> ```js
> {
>   "stm32cubeide":"/opt/stm32cubeide/ ",
>   "stm32cubeprogrammer": "/opt/stm32cubeprog/"
> }
> ```       
>   
> Then I think it is better to use these two parameters, and it looks neater. Then cortex-debug finds the things needed in it.
> Only configure to the top-level directory of the software installation.
> ```

@yueduz While I agree it looks neater, there are good reasons to stick to the current configuration method:

1) I recommend not to add any changes which make the plugin incompatible with existing settings. People hate it when an upgrade breaks things. Instead of a new setting `stm32cubeide` please use the existing `serverpath` and `armToolchainPath` which has been around for years and people are well accustomed to it. There is absolutely no need for another setting.

2) The shared path can be derived from either `serverpath`  or `stm32cubeprogrammer`. I believe (have not tested it) `stm32cubeprogrammer` path should be sufficient to resolve the shared library path. Please try this.

3) Auto discovery should still work. So ideally neither `serverpath` nor `stm32cubeprogrammer` should be required if the ST tools are installed correctly using the default location.

Good luck with the implementation and thank you for doing this work.

yueduz:
Today, I spent almost half a day studying how to reframe the parameters of stlink. Since everyone does not like to change, then I will not change the parameters. The changed code has been requested to merge

yueduz:
I saw in the code that the default parameters of stlink depend on the installation directories of STM32CubeProgrammer and STM32CubeIDE at the same time. However, the key components of STM32CubeProgrammer are provided in the STM32CubeIDE installation package under Linux. Therefore, only STM32CubeIDE needs to be installed under linux to satisfy the dependency. I don't know windows and macos. Is it necessary to remove the dependency on STM32CubeProgrammer?

yueduz:
```json
       {
            "type": "cortex-debug",
            "request": "launch",
            "servertype": "stlink",
            "cwd": "${workspaceRoot}",
            "executable": "${workspaceRoot}/build/stm32wle5x_lora.elf",
            "name": "Debug (STlink)",
            "device": "STM32WLE5CB",
            "runToEntryPoint": "main",
            "stlinkPath": "/opt/stm32cubeide/plugins/com.st.stm32cube.ide.mcu.externaltools.stlink-gdb-server.linux64_2.0.100.202109301221/tools/bin/ST-LINK_gdbserver",
            "stm32cubeprogrammer": "/opt/stm32cubeprog/bin",
```
The above configuration allows me to work. Because the default installation directory of the Linux distribution I use is different from the cortex-debug default parameters, and the parameters are configured to the last level of the directory, I have to modify the configuration every time I update the stm32cubeide. If you provide a parameter configuration to the stm32cubeide installation directory, then I don't have to modify the configuration every time I upgrade. But this will add new parameters.

haneefdm:
I don't like additional parameters. Or even adding stuff like LD_LIBRARY_PATH.

Why isn't LD_LIBRARY_PATH already in your shell environment setup through you ~/.bashrc or equivalent? Please make do with `serverpath`. Every item you add to the package.json creates a WHOLE lot of work for me.

PR #378, opened 2021-01-18
Update external.ts

v-rakkun:
When application loaded to RAM memory, Reset after target download will clear the RAM memory. Because of this, debug will not be happen.

### DIFF
--- a/src/external.ts
+++ b/src/external.ts
@@ -37,7 +37,6 @@ export class ExternalServerController extends EventEmitter implements GDBServerC
         const commands = [
             'interpreter-exec console "monitor reset halt"',
             'target-download',
-            'interpreter-exec console "monitor reset halt"',
             'enable-pretty-printing'
         ];
 


v-rakkun:
@Marus, Please provide your feedback.

haneefdm:
This may be a breaking change for some users. Did you try the override option to customize the startup?

v-rakkun:
Hi Haneef,

Can you please provide override option details, how we can use that?

Thank you
Rakesh K

Get Outlook for Android<https://aka.ms/ghei36>

________________________________

v-rakkun:
I haven't tried with override option. Can you please give override option usage to ignore reset after target download if possible.

haneefdm:
Open launch.json start typing  `override` and you will see a whole bunch of them. You can customize a lot of things

Marus:
Yeah - I don't like the looks of this as it currently stands - as @haneefdm mentioned I'd worry about backwards compatibility issues, as well as just general compatibility with certain GDB servers (at least some servers/devices end up in a weird state after the flash - so not doing the reset can cause issues).

I expect if you put the something similar to the following in your launch.json it will solve the issues with your use case (assuming the target doesn't start running automatically after target-download; if it does you could try adding `'interpreter-exec console "monitor reset halt"'` after the `target-download` but you probably won't end up being right at the start of your program unfortunately)

```
...
     overrideLaunchCommands: [
             'interpreter-exec console "monitor reset halt"',
            'target-download',
            'enable-pretty-printing'
     ],
...
```

Marus:
Since I think you can achieve the result you are wanting without changes to the extension - I'm going to close the PR for now - if you cannot solve it with the facilities currently provided please feel free to re-open and we can investigate further.

PR #381, opened 2021-01-19
icons: use vscode built-in icons

manuargue:
In this way, icons will evolve along with vscode and support
for all the themes color will always work. Also it reduces
a bit the extension size.

with dark theme

with light theme


Marus:
Looks good - thanks - will include in the next release.

manuargue:
> Looks good - thanks - will include in the next release.

That's great :) btw when is it planned next release?

PR #384, opened 2021-01-20
bugfix: handle cancel in set format command

manuargue:
Abort to set the peripheral format if the quickpick result is undefined.

Before the fix:
Screenshot_2021-01-21_01-35-31


Marus:
Thanks for the fix. Will include in next release.

Issue #370, opened 2020-12-17
Remember scroll position in Cortex Peripheral panels


timblakely:
Hopefully this is an easy one for ya :) Love that the Peripherals panel remembers which peripherals/registers are expanded, but it seems whenever I restart debugging the panel's scroll is reset back to the top (possibly after re-parsing the SVD?). This makes debugging peripherals towards the end of the alphabet like `TIMx`, `WWDG`, etc a bit of a chore since it requires re-scrolling down the peripherals panel to find it again. It would be ideal if the panel could remember the last scrolled position and reposition itself there upon next debugging session.

Marus:
Unfortunately, while this would be very nice, I don't think this will with the current extension APIs that I have for visual studio code - it is something I will look into though.

manuargue:
@timblakely As a workaround for avoiding scrolling down to the bottom, you can focus on the peripheral view and start typing the peripheral name for searching it.

Probably it would be nice to have some sort of "pinning" button on each peripheral so that it can be brought to the top until unpinned.

manuargue:
@Marus I could add a "pin" button on each peripheral node so that it stick to the top of the peripheral list. This state can be persisted. It's handy when debugging the same peripheral frequently. What do you think? I can work on this.

Marus:
@manuargue - Pin/Unpin at the peripheral layer, which would just sort those ones to the top, I think is a great idea, and I think would help a lot with what @timblakely actually wants to accomplish. Shouldn't be too hard to actually achieve either I don't believe.

If you want to take a stab at this I'd be happy to look at it when ready and include it in the releases.

timblakely:
Absolutely, the ability to pin a device to the top of the list would solve it completely, and likely be even more useful when working in situations involving multiple subsystems (DMA, interrupts, etc)!

manuargue:
Great, I'll work on this and try to have the pr opened by this weekend.

timblakely:
Works like a charm, thank ya much! Unless you've got additional work to do, feel free to close this bug.

Marus:
Closing this since the alternative of pinning was released. Thanks again for the PR @manuargue

PR #385, opened 2021-01-21
issue #370: add toggle pin button to peripherals

manuargue:
This solves #370. It allows to pin a peripheral node to the top of the list for quicker access. Useful when debugging some peripheral to avoid scrolling or searching it every time the debug session is restarted. The state is persisted across debugging sessions.

Demo (I restarted the debugging session to show the state is persisted)
pin-peripherals

Although there is a **TODO**. In order to use `ThemeIcon` is needed to update `vscode` dependency to use `@types/vscode` as described [here](https://code.visualstudio.com/updates/v1_36#_splitting-vscode-package-into-typesvscode-and-vscodetest). I tried this already (removing `vscode` and installing `@types/vscode@1.34.0` to keep the same API as current) but I believe the tests need to be updated as well. The issue with `ThemeIcon` is that in current dependency, the constructor is private, but somehow in the types version (also v1.34) is public.

I can address this update in a separate pr and merge this one afterwards. But I may need some quick support from you @Marus in order to migrate the tests as well, if it's needed.

### DIFF
--- a/package.json
+++ b/package.json
@@ -132,9 +132,14 @@
                 "title": "Refresh"
             },
             {
-                "command": "cortex-debug.peripherals.togglePin",
-                "title": "Toggle Pin",
+                "command": "cortex-debug.peripherals.pin",
+                "title": "Pin",
                 "icon": "$(pin)"
+            },
+            {
+                "command": "cortex-debug.peripherals.unpin",
+                "title": "Unpin",
+                "icon": "$(pinned)"
             },
             {
                 "category": "Cortex-Debug",
@@ -1453,7 +1458,11 @@
                     "when": "false"
                 },
                 {
-                    "command": "cortex-debug.peripherals.togglePin",
+                    "command": "cortex-debug.peripherals.pin",
+                    "when": "false"
+                },
+                {
+                    "command": "cortex-debug.peripherals.unpin",
                     "when": "false"
                 },
                 {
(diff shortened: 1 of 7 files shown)


RisinT96:
Damn looks great!

I think if you put the "pinned" icon on the right (and make it the un-pin button) would be even better.

manuargue:
> Damn looks great!
> 
> I think if you put the "pinned" icon on the right (and make it the un-pin button) would be even better.

I tried a few things but couldn't make it work. The TreeItem only allows to have the icon before the label. Perhaps having two separate commands defined in package.json, one for pinning and another for unpinning, with different icons that are shown mutually exclusive according to some context. @Marus any suggestion?

Marus:
Yeah - I do think it is off to a great start. Regarding the icon - yeah - I don't know off the top of my head a good way to change it depending on the context - other than what you had thought of perhaps two commands that with different icons that you only show one of the two.

I'm planning on doing the v0.3.8 release tonight or tomorrow morning (just have a few more scenarios I want to test out) - and can hopefully look at this on the weekend to go in a v0.3.9; don't know exactly when that will go out - but hopefully won't be too much of a delay - as I also want to get some of my other things WIP things out the door.

Marus:
@manuargue - Regarding the ThemeIcon and switch from `vscode` to `@types/vscode` - that does seem like a good plan - but maybe I'll try to tackle that this weekend - as I'm also planning on updating the vscode runtime version from the rather ancient 1.34 that it is currently using (to unlock me using new features) - and want to update a bunch of other dependencies. Will let you know once I've done that - and you will be able to update this PR.

manuargue:
I've separated the pin/unpin commands so that only one of them is shown at a time, by creating a new context `peripheral.pinned` for the items pinned. Beware that commands that affect pinned and unpinned peripherals show match both in the `when` clause (e.g. using `viewItem =~ /peripheral.*/`).

This update remove the dependency on `@types/vscode` because the tree item icon is no longer needed. So is ready to merge and hopefully can be included in this release, if no further comments :relaxed: 

Demo:
pin-peripherals-2

Marus:
This looks really good. In some ways I still liked the tree icon - and will probably bring that back later - just like how that icon is always visible. But great work on this, and will try to get it out to the public soon.

manuargue:
All right, I’m glad to hear that. So I’ll leave it up to you to merge it whenever you consider appropriate. I also like the icon always visible, we can bring it back after updating vscode.

Marus:
So I've merged this in and it will go out with the v0.3.9 release - which hopefully can happen sometime early next week.

PR #391, opened 2021-01-31
bugfix: complete launch even if failed to break on main

manuargue:
When `runToMain` is enabled and it fails to set the temporary breakpoint (e.g. function main does not exist), complete the launch as if the breakpoint had not being defined to avoid hanging the debug session.

Rationale: some applications may name the entry point different than `main`. In such cases, the user will be notified with a warning message but the debugging session will carry on as normal.


Marus:
Good fix

PR #392, opened 2021-01-31
issue #389: replace runToMain with runToEntryPoint

manuargue:
Replace runToMain with two configuration options: runToEntryPoint
(default to false) and entryPointName (default to main).

runToMain become a deprecated configuration option which would set those two
if it was set and also the newer options not given. If both runToMain
and runToEntryPoint are enabled, runToEntryPoint takes precedence.

### DIFF
--- a/src/gdb.ts
+++ b/src/gdb.ts
@@ -395,7 +395,7 @@ export class GDBDebugSession extends DebugSession {
                     this.handleMsg('log', dbgMsg);
                 }
 
-                this.disableSendStoppedEvents = (!attach && this.args.runToMain) ? true : false;
+                this.disableSendStoppedEvents = (!attach && this.args.runToEntryPoint) ? true : false;
                 this.miDebugger.connect(this.args.cwd, this.args.executable, commands).then(() => {
                     this.started = true;
                     this.serverController.debuggerLaunchCompleted();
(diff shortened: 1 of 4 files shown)


src/gdb.ts
```diff
@@ -422,8 +424,13 @@ export class GDBDebugSession extends DebugSession {
                         }
                     };
 
-                    if (this.args.runToMain) {
-                        this.miDebugger.sendCommand('break-insert -t --function main').then(() => {
+                    if (this.args.runToEntryPoint) {
+                        let entryPointName = this.args.entryPointName || DEFAULT_ENTRY_POINT_NAME;
```
manuargue, line 428: When entryPointName is not specified, I’m not sure whether is better to silently set the breakpoint to the default value (e.g. by setting entryPointName default value in the config provider) or to let the user know that the entry point name can be defined and fallback to the default value (current implementation)

Marus: I don't think I'd bother with the message myself; that way it can largely function as a direct replacement for `runToMain` without having, just with the extra configurability for if they want a different name.

Myself I also think it's probably a bit cleaner to just set the `entryPointName` value to `main` in the config provider if it isn't already set.

manuargue: Updated to set default value of `entryPointName` to `main` in config provider

haneefdm:
Why not just have just one item where `runToEntryPoint` is a string type with value being the function name? Just a thought

Marus:
@haneefdm while the bool + string does keep it inline with the old deprecated property a bit more, guess there is no real reason why it couldn’t just be one setting where if it is null or empty (after trimming any white space) string it is treated as false and not used - while an actual value enables the function and sets the name. Thoughts on that @manuargue

manuargue:
I agree with @haneefdm idea, we can make it simpler and reduce the number of config parameters by keeping only `runToEntryPoint`. I pushed the new changes, pls help me review it.

Marus:
Going to merge this in to include in the next release (hopefully tonight or tomorrow morning). Any final tweaks I'll make. Thanks again @manuargue

Issue #364, opened 2020-12-01
Feature request: Improved variables view


KrestenHelstrup:
Hi
I have three feature requests for the variables view.
1) Allow us to format numbers between hex/dec.
2) If you select copy value on an array it should copy all members of the array for easy extraction of data.
3) Sort globals by name.

Thanks in advance

791314247:
顶，强烈要能选hex

haneefdm:
You can do this in the Watch window. You do `expr,x` for hex, `expr,b` for binary etc. Pretty much what gdb supports.

Variables window UI is controlled by VSCode but they recently added a feature to control the format but it is unclear how it works across data types and stack frames.

Sorting Globals by name is possible

KrestenHelstrup:
Hi haneefdm
I saw the answers from weinand in this thread : https://github.com/Microsoft/vscode/issues/28025
It let me to believe that cortex-debug must implement the formatting somehow. Unfortunately, I'm an embedded developer so I probably can't contribute much to the development.

Do you mean that sorting globals are already a possibility or it is possible to update cortex-debug to do it ?

Thanks in advance

bigbrett:
HUGE +1 for this issue. Seems like VSCode supports debuggers implementing this formatting now. This is one of the last features required to make VSCode the best embedded IDE out there!

KrestenHelstrup:
@Marus - Is this something that you would consider looking into ?
Alternatively, do you think we could find something in another debug extension we can port ?
Thanks in advance

haneefdm:
@KrestenHelstrup I have worked with weinand before. While he added that new feature, we have an issue as well.

We do not interpret values. We just report them as GDB does. Things can get messy and prone to error. Esp. floating point. It should be possible.

One thing that is not clear to me is how to remember which variable has which format between frames and every time you get a breakpoint

KrestenHelstrup:
@haneefdm Would it make sense to Weinand for some input ? Maybe he can point in the right direction. 
I'm up for contributing with testing/debugging, but my coding contributions would probably be limited as I'm an embedded C developer.

sidking:
Please, please add hexadecimal display for variables in VS Code.  Love the Marus extension and VS Code; adding this would make every embedded developer so much more productive (and happy).

KrestenHelstrup:
[...]

Issue #214, opened 2019-10-16
Feature request: Change core registers to be displayed hexdecimal


sergioagp:
Please show the rXX in hexadecimal or at least give the option to change between formats.
I stopped using cortex-debug because of this issue. I would like to use it again.

haneefdm:
For now, could you try the watch window. You can do `$r3,x` and it will display `r3` in hex. The register window uses the natural format (as specified by ARM/gdb) which does a better job, especially when dealing with floating point registers.

marduino:
@haneefdm 
Sorry for jump in.

I have the same issue and I would agree with sergioagp, in most cases it makes no sense to display non-hex format for registers, as they normally read together with dis-assembly code.

Would be good at least provide an option to allow select hex.

Thanks!

haneefdm:
@marduino No problem. Agree about an option.

The format used is the (natural) format recommended by ARM. Things like sp, pc, lr are hex, lower rXX are supposed to be general-purpose integer variables, etc. When it comes to floating point registers (sXX, dXX)  though, it becomes super important. You don't want to see them in hex. Granted most Cortex users are not interested but for those who do, it is nearly impossible to convert on the go.

I will do what I can to help implement with permission from @Marus ANd, there are limits for what we can do with the VSCode API and still being performant

marduino:
Thanks @haneefdm , I just saw the another thread about this, 
FYI. https://github.com/microsoft/vscode/issues/28025
I tried the command "set output-radix 16" from gdb terminal, it helps a lot.

Aerijo:
Version 0.2.7 allows individual configuration on right clicking the register, so it seems to have been removed at some point. Shouldn't it be possible to support this, and add an additional option "as returned by GDB" to the Auto / Hex / Decimal / Binary list? Or was this a performance problem?

haneefdm:
In the next release, you can have an option to display everything in hex or natural. Details in ChangeLog.

Per register config? Maybe. Stay tuned. A bit of work to get it right.

biovoid:
It's very important to be able to setup default view of core registers and variables. Also current UI of registers view requires a bit visual alignment if possible.

PR #395, opened 2021-02-06
WIP: Allow setting of an "override format" for non-peripheral registers

paked:
Hi!

I'm keen to get the formatting options back into the register view--it seems like this feature has disappeared since I last used cortex-debug (as discussed in #214).

What I've got so far is quite simple, and is basically the solution I came up that involved the least amount of refactoring. The main problem with this method IMO is the duplication between RegisterNode and PeripheralRegisterNode (especially the handlers for the relevant format commands in `extension.ts`) (and given the feature request for this same functionality in the variable view in #364, it seems likely it may be duplicated again).

Provided we clean up this code, is this something that could be merged in? I'd be happy to champion it, since I need this functionality anyway.


KrestenHelstrup:
Huge +1 here - which makes sense as I'm the author of #364 
In all other IDE's I have tried both registers and variables can be formatted with a simple right click and a menu item like "view as" then you can select between hex, decimal, binary.

PR #397, opened 2021-02-06
Fix argument for selecting JTAG or SWD interface for P&E Micro GDB server.

adamkulpa:
The Argument "-interface" for P&E Micro GDB server pegdbserver_console is used incorrectly.
This argument is used to select a manufacturer's hardware interface such as USBMULTILINK, CYCLONE, TRACELINK, OPENSDA.

However, in order to select debug interface no argument is needed for SWD, and the valid argument that selects jtag interface is: "-usejtag".

This change is necessary to be consistent with the configuration field "interface" in launch.json, which is an enumeration between the strings: 'swd' or 'jtag'.

The handling of the GDB P&E Micro server argument "-interface" and "-port" should be done with specific fields.

### DIFF
--- a/src/pemicro.ts
+++ b/src/pemicro.ts
@@ -100,8 +100,8 @@ export class PEServerController extends EventEmitter implements GDBServerControl
             serverargs.push(`-kernal=${this.args.rtos}`);
         }
 
-        if (this.args.interface) {
-            serverargs.push(`-interface=${this.args.interface}`);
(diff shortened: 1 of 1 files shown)


Marus:
Since I don't actually have a P&E debugger I'll have to assume you have the correct information here and I'll merge this to include in the next release.

Marus:
@adamkulpa - FYI - I've released this with V0.3.12

Issue #409, opened 2021-03-09
Need required settings for hex format view in watch and variables section


v-rakkun:
Hello Team,

I using cortex-debug extension to debug RT6xx series processor. It is working fine for debugging. In watch able to see the varibales in hex format by adding ,x to the variable name. Instead of adding it to every variable, is there any global setting to see variables and watch window values in HEX format by default ?

Thank you
Rakesh K

Marus:
No - for watch variables they are displayed in the GDB default format for the type of data; there is currently no global setting to override this.

bhargavshah1988:
@Marus what are your thoughts to add a configuration for this ?

FabianoGK:
You could add `set output-radix 16` to the `postLaunchCommands`.

v-rakkun:
Thank you @FabianoGK.

Post launch command(set output-radix 16) is working as expected.

PR #416, opened 2021-03-17
Added variable format support

v-rakkun:
Added support to see the values of variables in different formats
1. hexadecimal
2. decimal
3. octal
4. binary
5. natural format

### DIFF
--- a/src/backend/mi2/mi2.ts
+++ b/src/backend/mi2/mi2.ts
@@ -556,7 +556,7 @@ export class MI2 extends EventEmitter implements IBackend {
         x: 'hexadecimal'
     };
 
-    public async varCreate(expression: string, name: string = '-', fmt: string = 'h', scope: string = '@'): Promise<VariableObject> {
+    public async varCreate(expression: string, name: string = '-', fmt: string = 'n', scope: string = '@'): Promise<VariableObject> {
         if (trace) {
             this.log('stderr', 'varCreate');
         }
--- a/src/backend/mi2/mi2.ts
+++ b/src/backend/mi2/mi2.ts
@@ -556,11 +556,10 @@ export class MI2 extends EventEmitter implements IBackend {
         x: 'hexadecimal'
     };
 
-    public async varCreate(expression: string, name: string = '-', scope: string = '@'): Promise<VariableObject> {
+    public async varCreate(expression: string, name: string = '-', fmt: string = 'h', scope: string = '@'): Promise<VariableObject> {
         if (trace) {
             this.log('stderr', 'varCreate');
         }
-        let fmt = null;
         expression = expression.trim();
         if (/,[bdhonx]$/i.test(expression)) {
             fmt = expression.substring(expression.length - 1).toLocaleLowerCase();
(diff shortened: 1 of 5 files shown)


v-rakkun:
#409

haneefdm:
Hmmm, does that mean everything is displayed as one of these formats? So, if I have a char or a float I will no longer see it in its natural format? This not only affects the Watch window but Variable window as well, maybe even Hover, doesn't it?

This is also an all-or-nothing approach. I am not sure people will like this. No debugger does this AFAIK. Default should be 'n'. You can always override it for yourself. Don't do it for the rest of the users.

Your `package.json` should be improved. There should be an enum. See `serverType`. While you provide help, I could type anything there.

The varCreate is now twice as slow even for naturals. Since the format will always have a default. Sending commands and reading back the results is very expensive. Every time the program stops, the windows will take twice as long to update. We try to keep things performant.

I am sure @Marus will chime in but I don't like this PR. VSCode apparently added a new feature that we should explore.

v-rakkun:
It will support natural format also.

haneefdm:
But that is not the default and you have not addressed performance

haneefdm:
Here is the spec that guides all debuggers
https://microsoft.github.io/debug-adapter-protocol/specification

I can't remember. Some here pointed that out, but I can't find who said it or in the DAP spec. There is a Value format but it only supports hex. However, see

https://github.com/microsoft/vscode/issues/70377#issuecomment-686358142

and keep scrolling up and down. I haven't seen the new API, but if this works for us, it would be nice

haneefdm:
@v-rakkun That is not what I meant for `package.json`. See example below

```JSON
                            "servertype": {
                                "type": "string",
                                "description": "GDB Server type - supported types are jlink, openocd, pyocd, pe, stlink, stutil, qemu and external",
                                "enum": [
                                    "jlink",
                                    "openocd",
                                    "pyocd",
                                    "stutil",
                                    "stlink",
                                    "bmp",
                                    "pe",
                                    "qemu",
                                    "external"
                                ]
                            },
```
This prevents user from entering garbage and causing a crash at some point. See if this works for us

v-rakkun:
@haneefdm @Marus 

I have updated configuration using enums and extension will work as with existing functionality(without effecting performance). 

User can change the format of the variables window if required.

src/backend/mi2/mi2.ts
```diff
@@ -556,22 +556,22 @@ export class MI2 extends EventEmitter implements IBackend {
...
         }
         expression = expression.replace(/"/g, '\\"');
 
         const createResp = await this.sendCommand(`var-create ${name} ${scope} "${expression}"`);
         let overrideVal = null;
         if (fmt && name !== '-') {
-            const formatResp = await this.sendCommand(`var-set-format ${name} ${MI2.FORMAT_SPEC_MAP[fmt]}`);
+            const formatResp = await this.sendCommand(`var-set-format ${name} ${fmt}`);
             overrideVal = formatResp.result('value');
         }
```
v-rakkun, line 576: By default fmt is null, it won't effect the performance. If user wants to change the format then it will be updated and  condition will be executed.

v-rakkun:
By adding set output-radix 16 to the postLaunchCommands, seeing veriables in hex format.

PR #424, opened 2021-04-14
Dumb python script to describe options

wpd:
This is a python script I put together today that produces a GitHub Markdown table describing all of the debug attributes provided by this plugin.  I was going to spend more time cleaning it up for you and trying to make you think I'm a better python programmer than I really am (and perhaps adding a comment or two along the way), but I already spent enough time on this this afternoon (when I should have been doing the work I get paid to do), and tonight I'd rather work on the other stuff I want to play with (like debugging Cortex-M code with VSCode).  Perhaps you'll find this useful.  Perhaps not.  But you can have it and do whatever you want with it.

### DIFF
--- a/options-doc.py
+++ b/options-doc.py
@@ -1,0 +1,71 @@
+#!/usr/bin/env python3
+
+import pprint
+import json
+import re
+
+pp = pprint.PrettyPrinter()
+
+
+def get_properties(pkg, type):
+    attributes = pkg['contributes']['debuggers'][0]['configurationAttributes'][type]
+    return dict([(p, attributes['properties'][p].get('description', '(unknown)'))
+                 for p in attributes['properties']])
+
+
+with open('package.json') as f:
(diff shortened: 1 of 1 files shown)


haneefdm:
I like it. There are a couple of minor issues, it crashes on the last line of the script if the attribute does not exist in `all_properties` (example 'request' which is a reserved property of VSCode, but we needed it to see what kind of session it is).

I can put this in the source tree and make the output part of the release.

I have a lot more extra properties than what you were expecting in the script. There will always be differences

Unfortunately, I am not as skilled with python. Once I release it, I have to maintain/enhance it. But, I like and thanks

haneefdm:
It is merged into the 'rtt' branch that I will merge into master soon. It needs automation though so that it is part of the build process and should become a bit more robust and accurate -- especially handle properties that are defined for all debuggers. Thanks for the script

PR #441, opened 2021-06-07
Implements socket- or serial-carried SWO source for external connections

dmitry-tarnyagin:
Signed-off-by: Dmitry Tarnyagin <dt@jetro.no>

### DIFF
--- a/src/common.ts
+++ b/src/common.ts
@@ -50,7 +50,8 @@ export class SWOConfigureEvent extends Event implements DebugProtocol.Event {
     public body: {
         type: string,
         port: number,
-        path: string
+        path: string,
+        host: string
     };
     public event: string;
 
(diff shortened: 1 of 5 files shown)


dmitry-tarnyagin:
Hi,

The commit implements SWO over TCP/IP or serial for the "external" connection type. It is useful for remote debugging with port forwarding over SSH.

Best wishes,
Dmitry

haneefdm:
I am sorry to say that we had already merged such a feature. It is now in the main-line. I initially thought it was the same as yours but it looks like a different fix for the same issue.

Main difference is that the current method uses `[host:]port` syntax rather than separate host and port fields.

Can you see if what is on `master` satisfies your PR as well? As is I cannot merge this PR. Yours may be better but some one had given me their implementation a while ago

haneefdm:
We have released support for SWO. Can you review that functionality vs yours and see if anything should be added. I cannot easily resolve your changes. Yours seems just as good. The main difference I see is you use separate host and port fields vs what is on master overloads the 'port' to also specify a host.

Sory, but I have to close this PR in favor of what is already there. If you would like to reopen or have further improvements, please feel free.

PR #447, opened 2021-06-21
SWO showOnStartup Fix

pearson:
The current cortex-debug has an SWO parameter "showOnStartup" that is intended to flag for the "panel" view to be switched to the SWO output. However, this does not currently work. If the "panel" is switched to the SWO output immediately after it is created, an unchanging black view will appear. Manually switching the view to another and back will then show the SWO output. The reason for the unchanging back view is unknown.

This fix works-around the issue by adding a timer to the SWO console startup code. One second after the console is started the timer will then switch the "panel" to the SWO output view.

This work-around is a hack, but it was the only obvious fix.

### DIFF
--- a/src/frontend/swo/decoders/console.ts
+++ b/src/frontend/swo/decoders/console.ts
@@ -13,14 +13,19 @@ export class SWOConsoleProcessor implements SWODecoder {
     public readonly format: string = 'console';
     private port: number;
     private encoding: string;
+    private showOutputTimer: any = null;
     
     constructor(config: SWOConsoleDecoderConfig) {
         this.port = config.port;
         this.encoding = config.encoding || 'utf8';
         this.output = vscode.window.createOutputChannel(`SWO: ${config.label || ''} [port: ${this.port}, type: console]`);
 
-        if (config.showOnStartup) {
-            this.output.show(true);
+        // A work-around. A blank display will appear if the output is shown immediately 
+        if (config.showOnStartup) {       
+            this.showOutputTimer = setTimeout(() => {
+                this.output.show(true);
+                this.showOutputTimer = null;
+            }, 1000);
         }
     }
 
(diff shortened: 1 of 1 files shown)


haneefdm:
Hi @pearson , could you reduce the timeout. Like 100ms. How low can you go? One second is super long -- enough time to mess with user interactions

Did this always happen? I doubt @Marus would have released the code if this was the case. Wondering out loud, what is causing this.

pearson:
Hi @haneefdm , I'd wondered about whether this was a problem originally, too. My thoughts are probably similar to yours, in that it probably did work, and wondering what is causing the issue. However, the `showOnStartup` parameter isn't documented, and I'm not sure if there's meaning in that.

I changed the timer to 1ms and it does seem to work for me. It also works at 0.1ms, but it does not work with a timer of zero. 

For whatever reason, it does look like the `output.show(true)` simply needs to be done outside of the current context. However, my knowledge of VS Code extensions is still pretty limited, though, both for digging for the root cause and for creating a different fix.

Anyway, I am fairly convinced that there is some root cause to the issue, and that this fix is a hack. It works for me at the moment, though.

haneefdm:
Yes, thanks. Lets do 1ms for now. There are 3-4 other places where this same code exists, btw.

I have a feeling that createOutputChannel(...) hasn't actually finished creating the window yet when we called show and doing a setTimeout(...) allows pending tasks to be completed before the callback is called. That function should have returned a Promise/Thenable letting us know that it is a (partially) deferred operation.

pearson:
> There are 3-4 other places where this same code exists, btw.

That's good to know. I'm not quite comfortable with using timers for queuing operations, but then JavaScript has always been kind of odd.

> I have a feeling that createOutputChannel(...) hasn't actually finished creating the window yet when we called show and doing a setTimeout(...) allows pending tasks to be completed before the callback is called. 

That makes sense, although there doesn't seem to be any mention of the output channel being created asynchronously. The documentation suggests the output.show(true) can follow the createOutputChannel(...).

PR #453, opened 2021-07-10
Add support for parameter serialNumber when use stutil as servertype.

TangCheng:


src/stutil.ts
```diff
@@ -111,6 +111,10 @@ export class STUtilServerController extends EventEmitter implements GDBServerCon
             serverargs = serverargs.concat(this.args.serverArgs);
         }
 
+        if (this.args.serialNumber) {
```
haneefdm, line 114: Wrong place for this code. It should at line 109. The `if (this.args.serverArgs) {` should be the last block.

haneefdm (changes requested):

PR #457, opened 2021-07-28
Accept undefined stop reason

trond-snekvik:
In some cases, the GDB client will not include a reason field in the
async stopped event. Most notably, the SEGGER JLink server program entry
stop event only does in some corner cases, and as a result, most JLink
launch sessions start in the reset code with the exception error widget.

image

Additionally, as the GDB class relied on the erronous `stopped` event to
set `this.stopped`, the `setFunctionBreakpoints` request running after the
"Initialized" event will deadlock waiting for a "generic-stopped" event
that already occurred.

Add a `reason !== undefined` guard for the stopped event in MI2 to fall
back to a simple "generic-stopped" event for this scenario. Mark the
debugger as "stopped" when the debugger is ready to avoid deadlocking 
the configuration procedure without the invalid stop event.

### DIFF
--- a/src/backend/mi2/mi2.ts
+++ b/src/backend/mi2/mi2.ts
@@ -179,7 +179,7 @@ export class MI2 extends EventEmitter implements IBackend {
                                         this.log('stderr', 'Program exited with code ' + parsed.record('exit-code'));
                                         this.emit('exited-normally', parsed);
                                     }
-                                    else {
+                                    else if (reason !== undefined) {
                                         this.log('console', 'Not implemented stop reason (assuming exception): ' + reason);
                                         this.emit('stopped', parsed);
                                     }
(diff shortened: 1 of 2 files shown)


haneefdm:
HI @trond-snekvik , this may not be the best way to handle this. What you did appears to not send an event to vscode which means it will not query the stack and may not set the front end status. Real exceptions can occur after a program starts.

The problem here is that the gdb-server should have specified a reason for stopping. And maybe we should not be using. There is a stopped type called 'entry' and maybe we could use that once and just once.

If there is an issue with function breakpoint, we should also look at that.

haneefdm:
@trond-snekvik I made a change to handle the initial 'undefined' reason and changed the type to 'entry. Btw, this also happens with openocd. I cannot accept your change. It is on a branch called 'rtt'. Check it out and give any feedback.

About function breakpoints, we should file a separate issue with steps to reproduce it.

PR #484, opened 2021-09-17
minor comment cleanup

GitMensch:


haneefdm:
This is not a minor comment cleanup. You mentioned your issue in there, so I have to reject this PR.

Yes, my English is not good but that does not mean you inject other stuff into it.

GitMensch:
Sounds like a misunderstanding. Of course I can remove the issue, but _especially_ as it now also elaborates the reason _why_  objdump is used I'd actually suggest to keep it in.
If you want to I amend that issue reference out.

haneefdm:
Please remove the reference to the issue

Yes, I added those comments after being frustrated and to tell others who may walk by. With those changes, we were able to support C++ and Rust as well. Quite a few people using Rust!

GitMensch:
Done, replaced the issue reference with your note posted there.

PR #490, opened 2021-09-24
Breakpoints: Send responses for all requested breakpoints

trond-snekvik:
According to the DAP specification, the response to a Breakpoint
request should contain an entry for all requested breakpoints, including
those that won't be set.

Currently, the DAP server only returns entries for the breakpoints that
were successfully set, and returns an error if any of the break-insert
requests reported errors from the GDB server.

Change this to return unverified breakpoints for both erronous
break-inserts and ignored breakpoint requests. This makes the invalid
breakpoints show as greyed out in the UI, and removes the current error
toast for all debug setting. This is in line with the behavior of other
VS Code debuggers, like the NodeJS debugger.

Additionally fixes a bug in functionBreakpoints, where the debugger
returned the source parameter as a string, when it should be an object.

### DIFF
--- a/src/backend/mi2/mi2.ts
+++ b/src/backend/mi2/mi2.ts
@@ -438,7 +438,7 @@ export class MI2 extends EventEmitter implements IBackend {
                             if (result.resultRecords.resultClass === 'done') {
                                 resolve(breakpoint);
                             } else {
-                                resolve(null);
+                                reject(new MIError(result.result('msg') || 'Internal error', `Setting breakpoint condition`));
                             }
                         }, reject);
                     }
@@ -447,7 +447,7 @@ export class MI2 extends EventEmitter implements IBackend {
                     }
                 }
                 else {
-                    resolve(null);
+                    reject(new MIError(result.result('msg') || 'Internal error', `Setting breakpoint at ${location}`));
                 }
             }, reject);
         });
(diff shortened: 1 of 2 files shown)


trond-snekvik:
This issue was easiest to demonstrate by setting breakpoints in a file that wasn't included in the executable.
### Old behavior:
image
with the invalid breakpoint still shown in the UI like this: 
image

### New behavior:
image
and no error toast

haneefdm (approved):

haneefdm:
Thanks, these are really good changes. The DAP has evolved under us and every time I look at it, there is new stuff in there and some of it hasn't even been implemented by VSCode itself.

In fact, I was involved in a few additions to the DAP and I never got notified after it was released. There is no version history.

trond-snekvik:
Looks like they started following the language server protocol model, and made it into a standalone project: https://microsoft.github.io/debug-adapter-protocol/ 

Not sure when this change was made, but it's a bit better than the old way, at least. No idea how these releases are connected to VS Code's releases though. I'll keep an eye out for similar changes :slightly_smiling_face:

haneefdm:
They are not connected although it is the same people. DAP takes feedback from all debugger developers and do what makes sense. cpptools picks up some here and there depending on demand, priority, and resource availability.

Like the format support for watch variables. It is in the DAP but not in VSCode. We implemented our own more than two years ago -- suffix expr with `,x` for instance.

trond-snekvik:
Aw, that sucks. I see they added a menu contribution point for the variables view, but not the watch view, which is kind of dumb. I suppose this reduces their priority for adding the DAP version of this in VS Code too :/

PR #532, opened 2021-11-15
Allow value substitution for J-Link RTOS option

PhilippHaefele:
Allow the J-Link RTOS option to use https://code.visualstudio.com/docs/editor/variables-reference by moving the check to `resolveDebugConfigurationWithSubstitutedVariables()`

### DIFF
--- a/src/frontend/configprovider.ts
+++ b/src/frontend/configprovider.ts
@@ -4,6 +4,7 @@ import * as os from 'os';
 import { STLinkServerController } from './../stlink';
 import { GDBServerConsole } from './server_console';
 import { CortexDebugKeys } from '../common';
+import * as path from 'path';
 
 const OPENOCD_VALID_RTOS: string[] = ['eCos', 'ThreadX', 'FreeRTOS', 'ChibiOS', 'embKernel', 'mqx', 'uCOS-III', 'nuttx', 'auto'];
 const JLINK_VALID_RTOS: string[] = ['FreeRTOS', 'embOS', 'ChibiOS', 'Zephyr'];
@@ -21,7 +22,7 @@ export class CortexDebugConfigurationProvider implements vscode.DebugConfigurati
             return undefined;
         }
         config.gdbServerConsolePort = GDBServerConsole.BackendPort;
-        
+
         // Flatten the platform specific stuff as it is not done by VSCode at this point.
         switch (os.platform()) {
             case 'darwin': Object.assign(config, config.osx); delete config.osx; break;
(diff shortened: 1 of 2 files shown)


src/frontend/configprovider.ts
```diff
@@ -215,6 +231,40 @@ export class CortexDebugConfigurationProvider implements vscode.DebugConfigurati
...
+    private verifyJLinkConfigurationAfterSubstitution(folder: vscode.WorkspaceFolder | undefined, config: vscode.DebugConfiguration): string {
+
+        if (config.rtos) {
+            if (JLINK_VALID_RTOS.indexOf(config.rtos) === -1) {
+
+                let rtos_file_ext = '';
+                /* When we do not have a file extension use the default OS one for file check, as J-Link allows the parameter to be used without one. */
+                if ('' === path.extname(config.rtos)) {
+                    switch (os.platform()) {
+                        case 'darwin':
```
haneefdm, line 243: MacOS uses the `.dylib` extension`. Not `.so`

haneefdm (changes requested):

haneefdm:
I was wondering if `resolveDebugConfigurationWithSubstitutedVariables` should be used for all processing. I bet it did not exist in the API when this extension was created. But I have to scan line by line for what would break. I bet some stuff would not be needed like the following

https://github.com/Marus/cortex-debug/blob/3fc61810c47a833b86d865c23a64b286035c43c8/src/frontend/configprovider.ts#L26

I rarely edit this file.

src/frontend/configprovider.ts
PhilippHaefele: J-Link manaual (UM08001) states following `Specifies a RTOS plug-in (.DLL file for Windows, .SO file for Linux and Mac).` So I'm not 100% sure. Did you look into the installation folder? I haven't installed J-Link on my MAC and do not have it by hand right now.

haneefdm: Wouldn't you know it? JLink docs are wrong.
image

PhilippHaefele:
I was also thinking about that, but wasn't sure if everything then works like expected, as there're maybe parts that need replacement after processing (e.g. getting paths for STM32Cube stuff).
I found that function by accident after debugging several times and didn't understand where variable substitution was done.
With that knowledge it was very easy doing this change.
I can maybe test in the future, if at least most of the stuff is still working when doing the change, but would like to merge that one like it is for now.

src/frontend/configprovider.ts
PhilippHaefele: Fixed

PR #545, opened 2021-12-19
Set up dynamic link library for Stlink under Linux system

yueduz:


haneefdm:
@yueduz I cannot accept this change. You have modified the constructor for something OS specific.

I still think LD_LIBRARY_PATH should be set in your shell environment. Or replace.

Also, the code modified the adapter's environment. If you assign object 'a' to 'b' and then modify the contents of 'b', it is the same as modifying the contents of 'a'. This is what you did with 'env' and 'procEnv'

If I get a LD_LIBRARY_PATH issue, my first instinct is not to modify all others applications.

I will think about this for a bit but, no as is I cannot accept this change. Also, the code does not even pass TSLint (I know it is deprecated but that is we are still using).

And, this code only works for Linux. Meaning I have to go add support for other platforms?

CC: @hwmaier I am really confused as to why ST is becoming so complicated. We have the IDE, Programmer, GDBServer standalone, etc. What is the one version of the truth that I can hang my hat on?

haneefdm:
Sorry, @yueduz Thank you for making this effort. I know it is hard and a huge learning curve. I really appreciate it. It was just not the right way for me.

May I suggest, in the future, you start with creating a new issue instead of joining an existing closed one. We discuss solutions upfront and then proceed to create a pull request.

I should have paid more attention to **what** you were trying to do and instead, I was just showing you how to debug this extension when you needed help. Sorry for that.

yueduz:
If it looks like this, https://stackoverflow.com/questions/48595446/is-there-any-way-to-set-environment-variables-in-visual-studio-code
Adding the env parameter will work on all systems. Can you accept it?

hwmaier:
Hi @haneefdm, 

> CC: @hwmaier I am really confused as to why ST is becoming so complicated. We have the IDE, Programmer, GDBServer standalone, etc. What is the one version of the truth that I can hang my hat on?

The ST tools make it hard for the following reasons:

- Path location of tools depend on internal STM32CubeIDE version numbers which cannot be predicted (example 1.5.0.20201104092) . Result is every time you install a new STM32CubeIDE version, those path's change. Also those paths are deep and hard to resolve manually. Example: `c:/ST/STM32CubeIDE_1.5.0/STM32CubeIDE/plugins/com.st.stm32cube.ide.mcu.externaltools.stlink-gdb-server.win32_1.5.0.202011040924/tools/bin/ST-LINK_gdbserver.exe`
- ST's GDB server depends on the location of ST's flash programmer tool and that path must be passed as command line argument to it. So this adds some complexity as the tools are not independent and because of the issue above.
- Under Linux, ST's GDB server is not able to resolve the location of underlying shared library `libSTLinkUSBDriver.so`  if not started from the correct working directory. The shared library path also depends on the host processor platform (x86, x64, arm...). Refer to https://github.com/Marus/cortex-debug/pull/365#issuecomment-743924301

To make it easier for most users, I introduced an auto-discovery feature which configures the path's for the 3 configuration variables automatically using a default install location and regexes.

This works well for Windows, I believe it works for Linux as well with the exception of the shared library issue. Possible solution to the shared library issue are stated here: https://github.com/Marus/cortex-debug/pull/365#issuecomment-743942717 and yueduz tried to implement one of the approaches.

So from my point a user has two choices: Use the autodiscovery feature and install STM32CubeIDE to the default location or if more fine grained control of toolchain components is required, use the explicit settings `armToolchainPath`, `serverpath` and `stm32programmer`.

For power users not wanting to use the IDE or use different version of the tools or wanting the tools installed in different locations can use the explicit settings `armToolchainPath`, `serverpath` and `stm32programmer`.

From my understanding yueduz had two issues, firstly the installation of the cube IDE into a non-default path and the shared library issue. The first issue can be solved using explicit settings or by installing the IDE into the default location and the second using the environment injection approach or setting the LD_LIBRARY_PATH externally.

I believe solving the shared lib path properly would be beneficial for Linux users to have a smooth experience, so I think yueduz' contribution is useful.  I am not sure yet how best the platform issue is resolved and if setting the LD_LIBRARY_PATH is a better approach compared to changing the cwd.

hwmaier:
Another approach would be to remove the `stm32programmer` setting and the autodiscovery of `armToolchainPath`, `serverpath` alltogether and introduce an explicit `stm32idepath` setting as @yueduz has suggested which if present sets the 3 paths internally using the same regex based resolveCubePath() function.

This would be a breaking change but maybe it makes it clearer how cortex-debug is accessing the 3 components. It also offers better support to change the STM32CubeIDE versions as the top-level version number would be part of the setting:
```
{
            "type": "cortex-debug",
            "servertype": "stlink",
            "stm32idepath": "c:/ST/STM32CubeIDE_1.5.0/STM32CubeIDE",
...
```

Besides this being a breaking change the other disadvantage is that the location of the STM32CubeProgrammer cannot be set independent from the IDE but I wonder if that is really required. STM32CubeProgrammer is normally a stand-alone package.

hwmaier:
Some more infos, while for Windows there is a well-known installation path for STM32CubeIDE, this is different for Linux. If installed as sudo, it is `/opt/st/`, if installed as user it is `~/st`. Currently cortex-debug assumes a sudo install under `/opt/st/`.

sudo
```
STM32CubeIDE install dir? [/opt/st/stm32cubeide_1.8.0]
```
user
```
STM32CubeIDE install dir?  [/home/myusername/st/stm32cubeide_1.8.0]
```

haneefdm:
For sure can't change the cwd. That has other consequences for other servers and in the future for this maybe. We want to be consistent about this.

I don't like that implementation of the LD_LIBARY_PATH env. It was done in gdb.ts and mi2.ts besides modifying the adapters own env.

I will look into it. Both both Linux and Mac as well.

haneefdm:
@yueduz @hwmaier 
I just pushed changes to STLink (see ChangeLog). The linker paths should now work for both Linux and Mac. Did not test Windows but also did not touch it at least intentionally. A lot of changes had to be made to the discovery methods for Linux. If you have the time, please try it out, review my changes in the STLink controller.

Btw, the `reset halt` does not work for `stlink server`, so I changed it and you get normal looking messages from the server. We will have issues regardless for multi-core systems.

I did not do the `stm32idepath` way -- maybe another time (have to add 4 variables to package.json and still maintain all the old stuff).

Note that I have TONS of other changes mixed in the code I just pushed. No one is probably looking during the Holidays.

hwmaier:
@haneefdm I am currently on holidays and won't be able to fully test as I don't have an ST-Link device with me. But I tried starting the ST-LINK_gdbserver.exe and it appears to load correctly on Windows. So I think we are still OK with the Windows support.

haneefdm:
@hwmaier No worries have a great holiday. This stuff will be still here

haneefdm:
> If it looks like this, https://stackoverflow.com/questions/48595446/is-there-any-way-to-set-environment-variables-in-visual-studio-code Adding the env parameter will work on all systems. Can you accept it?

The env method works by default. It is already a feature of VSCode. You can use existing environment variables to define other stuff just like you do others.

https://code.visualstudio.com/docs/editor/debugging#_variable-substitution

But this is not used to create new ones through which make sense for native programs. So, no you cannot have properties like "env", "envFile", but you can do something like
```
    "serverpath": "${env:XYZ}/bin"    // XYZ is an environment variable.
```
Just like Cortex-Debug, other debuggers support many variables specific to their debuggers. Perhaps you can ask Microsoft, because this a generic feature.

haneefdm:
Since the support for this is already in the preview release, closing this one.

PR #548, opened 2021-12-22
Pass accessType from field into PeripheralFieldNode

MartinBeckerKT:
Hi there, 

we realized that for our SVD file the Cortex Peripherals tab could not create proper Tooltips and some could not be expanded because of an error.

It seems like the accessType was not passed over for fields.

This change works for us but is untested for other use cases!
 - Tooltips now show specific registers as (write-only).
 - Expanding Registers and Write to write-only fields also works

Please feel free to check and merge our PR.

Martin

### DIFF
--- a/src/frontend/svd.ts
+++ b/src/frontend/svd.ts
@@ -175,13 +175,17 @@ export class SVDParser {
                 }
             }
 
-            const baseOptions = {
+            const baseOptions: any = {
                 name: f.name[0],
                 description: description,
                 offset: offset,
                 width: width,
                 enumeration: valueMap
             };
+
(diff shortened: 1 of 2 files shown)


haneefdm:
I am not sure this is the right fix. I didn't write the code but the spec says that `access` does not have to be specified and is inherited from the parent. In the code I see, it can crash if access not specified? Am I reading your code right?

image

MartinBeckerKT:
Thanks  @haneefdm for your feedback!

You are right, I now consider the accessType optional. I just updated my branch and also tested with and without accessTypes in my svd.

PR #642, opened 2022-04-22
Add uC/OS-II support in RTOSView

PhilippHaefele:
@mayjs and i added suport for uC/OS-II

Therefore we also "refactored" some parts of the RTOS View code:

- There are now 4 different types of columns => Normal, Percentage, Link, State
- running row is not determined by a running boolean inside RTOSThreadInfo (formerly FreeRTOSThreadInfo)
- the RTOS specific part is now also able to pass additional CSS into a style element in the common html (also to overcome Content-Security-Policy issue with inline styles)
- Text in cells is now center aligned for better readability (percentage column type looked very ugly when right aligned)
- uC/OS-II RTOS View uses two optional launch parameters inside an element named "rtosViewConfig" => if this keeps being ok, we should add the (empty) root element to the package.json

grafik

Also did some spelling fixes. Sorry for that being part of this PR


PhilippHaefele:
Sorry but need to change something before this can be merged, but the review/feedback can start anyway

haneefdm:
Nice. I see that you got percentage bars going

If you use the CSS tags you can use the standardized colors everything and get proper right/left justification. Also use fixed width fonts. This CSS also deals with VSCode themes so it will be legible in all themes. If not, this is where I want changes made and not hard code them into our typescript code..

See file `resources/rtos.css` and the rtos.js file adjacent. The JS file has some junk from where I copied stuff from but it is mostly used to create a Help section at the bottom.

haneefdm:
Looks like you are already using that CSS but not sure why your fonts and justifications are different.

haneefdm:
@PhilippHaefele I know you are not done but, I am not even sure how to review this. There are far too many changes that are just white space and reformatting and some of it not good. For instance, there is a reason I used 
```
/*html*/`
 <somehtml>
    ...
</somehtml>
`
```
There is an extension that syntax highlights strings, caught some issues while I was typing. The comment at the start of the string tells the extension what language to expect https://marketplace.visualstudio.com/items?itemName=Tobermory.es6-string-html I found that MS folks use this extension, so I started using it.

I had tables of code (column aligned) that got re-formatted to the point that I cannot tell what the non-formatting changes are.

You probably ran some kind of code formatting tool ... I also use one but it does not do what it did to you.

PhilippHaefele:
> Looks like you are already using that CSS but not sure why your fonts and justifications are different.

If you're revfereing to the PEND MULTI => as we reused the help button stuff here, we also got the `font-size: 15px;` from there. Easy to copy that and remove the font-size setting there.
Did you mean that or something else?

> There is an extension that syntax highlights strings, caught some issues while I was typing. The comment at the start of the string tells the extension what language to expect https://marketplace.visualstudio.com/items?itemName=Tobermory.es6-string-html I found that MS folks use this extension, so I started using it.

Wasn't aware of that plugin at all. Did the html change (manually) to have nice html output (to easier analyze the output html) which is correctly intended (no native dedent in typescript). As the indentation is not really neccessary for the end result and the plugin brings in a need to have the whole string at a time, i will revert back to the original style.

I will add the es6-html and tslint extensions to the workspace extension recommendations, as it is very hard for others to guess what you're using. Any more extensions you use for development?

src/frontend/rtos/rtos-common.ts
```diff
@@ -60,9 +77,9 @@ export abstract class RTOSBase {
     // Refresh the RTOS structures
     public abstract refresh(frameId: number): Promise<void>;
 
-    // Return a string that represents the RTOS state. Ideally, it should return a grid/table that is
-    // hosted in an upper level structure
-    public abstract getHTML(): string;
+    // Return a string tuple (html + style element content) that represents the RTOS state.
+    // Ideally, it should return a grid/table that is hosted in an upper level structure
+    public abstract getHTML(): [string, string];
```
haneefdm, line 82: Perhaps an interface should be defined for return type like
```typescript
export interface HtmlInfo {
    html: string;
    css: string;
}
```

resources/rtos.css
```diff
@@ -1,6 +1,6 @@
 .threads-grid {
   gap: 0;
-  text-align: right;
+  text-align: center;
```
haneefdm, line 3: I am not comfortable with changing the default to 'center'. I want all numeric fields at least to be right-justified. Center justified looks weird. We should figure out which columns should be centered.

resources/rtos.css
```diff
@@ -20,18 +20,23 @@
...
 .running {
   color: var(--vscode-symbolIcon-classForeground);
-  font-weight:bold;
+  font-weight: bold;
+}
+
+.backgroundPercent {
+  background: linear-gradient(to right, var(--vscode-charts-blue) 0% var(--rtosview-percentage-active, 0%), rgba(0, 0, 0, 0) var(--rtosview-percentage-active, 0%) 100%);
+  background-clip: padding-box;
 }
```
haneefdm, line 41: Nice! I have to see how this works after I merge

haneefdm:
@PhilippHaefele,  @mayjs 

So, how hard was it to implement a new OS? I am sure more re-factoring could have been done.

Btw, I am working on a major update (will go into preview in a week or so) and hopefully, I will include this PR. It fundamentally changes the way the debug adapter works for EVERY request. The API as defined by VSCode (the spec for the protocol is fine) was wrong. We trusted that and were returning thinking that VSCode will wait for a Response (either normal or error). Nope, if we can't return a value immediately, it needs to be a promise. Or else, we can get duplicate calls in many places. Anything new I added in the last year or so, I was already returning Promises. But all the old stuff was not and there is quite a bit of that. This was a source of some unreproducible issues.

haneefdm:
Okay, I want to test these changes and see it in action? I am not familiar with this OS or how to get started. I typically use this

https://www.st.com/en/evaluation-tools/nucleo-f429zi.html

I have other boards but this is the simplest one. Is there a way you can give me your example for this board? It might be some work but we may have to do that. Especially if I have to change something later.

PhilippHaefele:
> Okay, I want to test these changes and see it in action? I am not familiar with this OS or how to get started. I typically use this
> https://www.st.com/en/evaluation-tools/nucleo-f429zi.html

Will see what i can do. Need to create a new example with the OpenSource (was released like this not long ago) variant from https://github.com/weston-embedded/uC-OS2, as i use a commercial licensed one in a modified company's project for testing right now.

I think i have a STM32F4 discovery board lying around that is similar enough to yours for testing everything.

Will most likely end in a mix of CMake + GCC + STM32 HAL + J-Link OB + VSCode project. Any wishes from your side regarding the example?

PhilippHaefele:
> So, how hard was it to implement a new OS? I am sure more re-factoring could have been done.

It was quite ok to do so (as we already added Thread awareness support for Segger GDB Server, we where already into that topic). We didn't expected to work with html directly (more something generic like what Segger did in Ozone https://www.segger.com/products/development-tools/ozone-j-link-debugger/technology/rtos-awareness/). I think that things like the column type and running row we added make some parts more clear now. As you said, documentation (in code) and the code itself may needs a little bit more polishing/refactoring.

We really liked the getVar... etc. stuff though 🙂 

So overall it's in a state which allows to add support for new RTOS systems, but still things open to improve it.

@mayjs Anything to add?

src/frontend/rtos/rtos-common.ts
PhilippHaefele: Already talked to @mayjs about that and we left it as a tuple and wait for your feedback. As you also came up with this idea, i will change it 🙂

mayjs:
> @mayjs Anything to add?

I agree, the API to access data from GDB was nice to use :slightly_smiling_face:.
One thing to improve is the missing documentation of that API, I needed to read some code to understand how the functions are supposed to be used.
Also, we needed to frequently use regular expressions to extract C-string values from `*-val` fields, some helper functions to interpret these return values as different kdatatypes would be appreciated. 

The output rendering is where I see the most room for improvement.
I think the new column types feature is already a step in the right direction, maybe it would also be possible to add stronger abstractions and make this feature easier to expand with custom types for RTOS's.
As Philipp also mentioned, it would be nice to reduce or get rid off manual HTML output in the RTOS implementation.

resources/rtos.css
PhilippHaefele: As most numbers are in hex or have the same/similar number of digits, i personally do not see a big difference in having it center aligned.
Also i'm not used to see right aligned stuff a lot (here in Germany).
Also the new percentage view eliminated a lot of numbers for a much better visual appearance (way to much "meaningless" address entries which needed manual calculations before) at least for the uC/OS-II implementation.
But should be easy with css to set the alignment per column.

My preferences are:
- Text columns should be either center or left aligned

- Percentage columns are center aligned

- not sure if Priority is a numeric value for you as in FreeRTOS it has a second value behind it. But no clear preferences from my side here

- id also not a clear preference on that

One thing to test here is, if mixing alignment in a table really looks good and allows easy reading of all elements.

haneefdm:
> I think i have a STM32F4 discovery board lying around that is similar enough to yours for testing everything.

I can order a Discovery board if it is not too expensive. Let me know which one. Or something else.

All in all too bad we had to do this ourselves. Microsoft could have open-sourced theirs and we could have added/contributed these OSes over there. They never responded to me or others.

https://github.com/microsoft/vscode-embedded-tools/issues/3#issuecomment-1067217250

When the above comment was posted, I had not even started on the RTOS stuff. It was just in my head.

> I agree, the API to access data from GDB was nice to use 🙂.

The implementation of this API will work with any debugger that follows the spec, not just Cortex-Debug. Technically, this whole RTOS thing can be a separate extension. Cortex-Debug has no idea what is going on here. ZERO special things were provided by Cortex-Debug.

Yes, I hated doing anything HTML/CSS as well. I am used to C++/C# Widgets. I used the MS kit here (still in Beta) but was all HTML/CSS/DOM. So is ALL of VSCode. Even the SWO/RTT graphs are like this.

haneefdm:
I have an update that will affect this merge. Please see

https://github.com/Marus/cortex-debug/commit/48f2bed2ecb1ac0b2d4b9c4c90403f2773236fca

Please merge these changes into your PR. Hopefully easy changes

We need to handle errors thrown by the debug adapter during discovery. This is happening now, with my changes there. We were doing this but not attempting retries. Should have always handled it! What is happening now is that the debugger briefly pauses at the reset vector to let VSCode update itself (should not be needed, bug in VSCode). It is just a few milliseconds but during that time, everyone monitoring the debugger (like the RTOS viewer) will get a paused notice -- meaning discovery starts and before or during even the first query can be made, the debugger has changed state to 'running'. This can also happen if the user single steps really fast -- queries get aborted all the time.

But, it was important for us to keep trying until and that is the change. A couple of lines in the freertos code had to be changed as well.

Note that the SPEC never calls for what exactly should happen for various kinds of errors. So, every debugger may do this differently. And neither does VSCode. It is all by experimentation.

PhilippHaefele:
> Btw, for FreeRTOS, I used a free 3rd party example from ST called `LwIP_HTTP_Server_Netconn_RTOS`

Would be nice, but that's not that easy. Will look into the project tomorrow and see what i can provide quickly.

> Maybe, there is such an example for uC/OS II as well. Something both of can use. Like I said, I will buy an ST board if needed (if it is not too expensive). I must have spent thousands of $$ already on Cortex-Debug :-)
> 

From our own experience it's quite difficult to get STM Nucleo boards right now. I would have bought yours too, if i could easily buy it somewhere.

> Curious, which device(s) do you all use?

I have following boards in hand quickly:

- STM32F4DISCOVERY (STM32F407VG)
- NUCLEO-F410RB
- NUCLEO-F413ZH
- NUCLEO-F767ZI
- NUCLEO-H723ZG

I may can ask some colleagues for more, but that's what i have at my place. Something you also have?

haneefdm:
>> Btw, for FreeRTOS, I used a free 3rd party example from ST called LwIP_HTTP_Server_Netconn_RTOS

> Would be nice, but that's not that easy. Will look into the project tomorrow and see what i can provide quickly.

Actually, it was one of STM32CubeIDE samples. I did not go out to get it. It is listed as 3rd party just like LwIP is 3rd party to them. I just searched in their IDE for freertos and picked one.

For a board, I have two criteria. 1) Must be in Stock and 2) Supported by JLink conv. utility

Looks like none of the boards listed are available at the distributor in the Americas. I found a few NUCLEO-F410RB at ST's website, but they are in Japan and they want $40 for shipping. Even ST has zero of the other boards.

haneefdm:
I noticed that the board I have is indeed is in stock ($23 USD) where as none of your boards seem to be in stock.

https://www.mouser.com/ProductDetail/STMicroelectronics/NUCLEO-F429ZI?qs=mKNKSX85ZJcE6FU0UkiXTA%3D%3D

Can you order one locally or do you want me to find something in Germany and ship it to you?

PhilippHaefele:
You're right, but 20€ for shipping is still a little bit too much.
But as i looked into the diffs between the STM32F429 and STM32F413 when implementing the example today, i think it should be easy enough for me to prepare everything without trying it out on the actual board.
Only very few files (startup file, cmsis-device file), clock settings and one define for the HAL will be different.

Got it compiling and some kind of running, but it stops working at some point... I will try to get it in a complete working state tomorrow.

haneefdm:
No hurry.

PhilippHaefele:
Did create a private GitHub repo (will go publich, once i have setup license information) with an inital project that should work for both STM32F413 and STM32F429. Was just the PendSV exception not being set to the right exception vector...
I will extend this project to show all different cases in a good manner.
Currently only one blinky task, but good enough to test for a first test.
Also will add toolchain fetching in some minutes.
Hopeing it works also well at your place.

PhilippHaefele:
> Looks like you are already using that CSS but not sure why your fonts and justifications are different.

Have fixed the collapse (fomer state) column type's style.
Looks much better now:
grafik

resources/rtos.css
PhilippHaefele: Jup, is a quite nice solution. Also experimented with mix-blend-mode to get the text inverted when inside the active part, but didn't got it working

PhilippHaefele:
@haneefdm I know you're busy getting the 1.5 stuff done, but would be nice to get a short feedback if we're on the right way in general (help, collapse style fixes etc.).

On top a statement in the `text-align` conversation would be fabulous, as i see this one as the only unclear point from my side right now.

haneefdm:
Very soon, @PhilippHaefele. The 1.5.x is a mess and I recruited some help.  Maybe a couple of days more.

I want to see this for myself before I comment. I still stand by my comments on justification -- such UI guidelines are even in books about data presentation. But I want to see this for myself.

haneefdm:
I don't understand what just happened. I just finished merging everything into my repo and then I see this.

https://github.com/Marus/cortex-debug/pull/642/commits/39f8c25ceca14d75ff973b312199d2aac05b21ca

With an important class (`ShouldRetry`) removed? This is being used elsewhere.

https://github.com/Marus/cortex-debug/pull/642/commits/39f8c25ceca14d75ff973b312199d2aac05b21ca

All these changes will go into 1.5.2 since 1.5.1 is already published.

PhilippHaefele:
Nice. Happy for any feedback 😃. I currently have one local error handling fix in a general part open (when trydetect promise fails at some point and therefore result or rtos is null/none due to some reason. Need to check why this happens) and it also happens quite often that detection of the RTOS fails because some variables of trydetect are not found (even if they are present). Restarting the debug session helps most of the time. Need to understand why this happens too. Maybe you can see/find something

PhilippHaefele:
Sorry @haneefdm haven't read your comments in the meanwhile. This function was double defined because i added it manually before and did some error when mergeing in master to this branch.
Sorry but i do not understand why you are doing this merge manually and not using the merge request. All the commits are gone which habe valuable information in it. Also the authorship is gone completely.
Is there a reason why you did it like this?

haneefdm:
>> Is there a reason why you did it like this?

Yes, because I needed see it before I committed the merge. I don't know how else to accept the PR just to take a look -- meaning a commit has already happened if I accepted a PR (after resolving the ChangeLog conflicts which I hate on github). And then I have to undo the entire commit. I am not a git expert and I have messed up many times.

How would have done it -- if you were scared of a whole bunch of changes that are not easily seen or understood? It was just too many changes for me.

Normally, if it is trivial, I don't do this. If it is large, then I have to use my own tools and not rely on GitHub diffs/merge.

Also, while I commited your changes and pushed, I am in the process of refactoring some RTOS related code. The FreeRTOS side now looks pretty bad as well.

haneefdm:
Btw, I would have squashed and merged anyway -- which I think is the github default for PRs

haneefdm:
Also tell me how to undo what I did with the merge so I can do it from github the normal way. But I need to reconcile the CHANGELOG.md file which I don't want to do. See the Current changelog on master.

haneefdm:
In the process of undoing the previous merge, I totally screwed it up. Fixing it now, not sure what I lost

PhilippHaefele:
> How would have done it -- if you were scared of a whole bunch of changes that are not easily seen or understood? It was just too many changes for me.

Add the fork as a second remote, checkout the branch an test it there. As you've seen you're also allowed to commit/push into the fork's branch of the PR.

> Btw, I would have squashed and merged anyway -- which I think is the github default for PRs

Wasn't aware of that, but yes you're right.

> Also tell me how to undo what I did with the merge so I can do it from github the normal way. But I need to reconcile the CHANGELOG.md file which I don't want to do. See the Current changelog on master.

We did manage it somehow. You could have done a git reset --hard YOUR_LAST_REF and then force push.
Just ask the user to merge/rebase upstream master and fix such things as the target version in the changelog.md (what i just did in the step of merging upstream master in rtos_view_ucos_ii_merge_fix).

> Also, I cannot merge your changes because of that last change of removing that class. Not even sure how it compile for you.

Sorry for that confusion of my upstream merge error. To clarify it. This was the state before removing it:
grafik

> Please fill me in on how to duplicate this. I can take a look at it as well. The class you removed was primarily to prevent that particular scenario. It happens when after issuing a request the debugger continues and thereby causing the query to fail but that just means we have to try again. It was not a total failure.

This https://github.com/Marus/cortex-debug/commit/d47bf81aca239d457618042e707e60f1d218b7c0 is the change i made prevent a exception which was something like "Can't read property status of undefined". Not sure if the error handling is ok. But as rtos is undefined we do not know which we should remove from `this.allRTOSes`, if I'm right

For the other issue, right now i can only say that i start a debug session (with a binary that has uC/OS-II) and get here:
grafik

grafik

It seems that it happens "all" the time on the first start after starting den VSCode [Extension Devlopemnt Host]. Maybe because the cache is not setup and some request take longer?

I also write this variable to 0 with `"postLaunchCommands": [ "set variable OSRunning = 0"]` to prevent misleading states on startup phase (RTOS is not initalized when halted at main). Could this result into a conflict?

PhilippHaefele:
Not sure if this is expected, but i only see this in DebugConsole when it is working:
`0000006285: 39-var-create  hover_bae4e7e41e2bf360232bceb4e34144807cf362ed372c379626163614e11c9b11 @ "OSRunning"`

It seems i need to check what happens inside `evaluateRequest` in `gdb.ts`, but it's already way to late to go to bed. Will look into it tomorrow/today

PhilippHaefele:
Ok it seems i found something:

grafik


`'MIError: Selected thread is running.\n    at errReport (D:\\Projekte\\cortex-debug\\dist\\debugadapter.js:12898:32)\n    at Object.<anonymous> (D:\\Projekte\\cortex-debug\\dist\\debugadapter.js:12922:21)\n    at D:\\Projekte\\cortex-debug\\dist\\debugadapter.js:12176:52\n    at Array.forEach (<anonymous>)\n    at MI2.onOutput (D:\\Projekte\\cortex-debug\\dist\\debugadapter.js:12151:15)\n    at MI2.stdout (D:\\Projekte\\cortex-debug\\dist\\debugadapter.js:12110:18)\n    at Socket.emit (node:events:526:28)\n    at addChunk (node:internal/streams/readable:315:12)\n    at readableAddChunk (node:internal/streams/readable:289:9)\n    at Socket.Readable.push (node:internal/streams/readable:228:10)'`

Tell me if i can supply more information for you (full debug output or some other values), but now it's really time to got to bed.
😴

haneefdm:
Yes, that is enough. Means between the time we started processing the request to the point we are trying to create a variable, program has continued. I can put a guard against that.

This is the problem with many async. things happening and there is no guaranteed order.

Issue #629, opened 2022-04-05
Chip stops running if I disconnect debugger with v1.4.x


goldwake:
I have noticed that with v1.4.x of the extension, I can no longer disconnect the debugger (hold Alt and press the stop button) and have the chip keep running with the debugger disconnected. It now just stops running as if I had hit the stop button.

When I revert back to v1.2.2 of the extension, it behaves as expected

EvilEli:
I experience the same issue (using v1.4.4)

haneefdm:
Which gdb-server? The detach/disconnect requests handled very differently (wrongly) by some gdb-servers. JLink vs STLink vs OpenOCD ...

@EvilEli Thanks for the command trace. I see that we are trying to detach followed by disconnect (which is what we should do) but I see a second disconnect and not sure where that is coming from.

goldwake:
my servertype is "openocd"

EvilEli:
I checked the output again -> does indeed disconnect twice

I also rolled back to version 1.2.2 which works properly for me as well.
Here is the output from that version:
```
[...]
0000065042: 35-exec-continue --thread 1
0000065054: GDB -> App: {"token":35,"outOfBandRecord":[],"resultRecords":{"resultClass":"running","results":[]}}
0000065055: GDB -> App: {"outOfBandRecord":[{"isStream":false,"type":"exec","asyncClass":"running","output":[["thread-id","all"]]}]}
0000065055: mi2.status = running
0000083049: 36-break-delete
0000083049: 37-target-detach
0000083050: GDB -> App: {"token":36,"outOfBandRecord":[],"resultRecords":{"resultClass":"done","results":[]}}
0000083051: GDB -> App: {"token":37,"outOfBandRecord":[],"resultRecords":{"resultClass":"error","results":[["msg","Cannot execute this command while the target is running[.\nUse]() the \"interrupt\" command to stop the target\nand then try again."]]}}
0000083052: 38-target-disconnect
0000083066: GDB -> App: {"outOfBandRecord":[{"isStream":false,"type":"notify","asyncClass":"thread-group-exited","output":[["id","i1"]]}]}
0000083067: GDB -> App: {"token":38,"outOfBandRecord":[],"resultRecords":{"resultClass":"done","results":[]}}
0000083067: -gdb-exit
0000083081: GDB -> App: {"outOfBandRecord":[],"resultRecords":{"resultClass":"exit","results":[]}}
```

EvilEli:
[...]

PR #656, opened 2022-05-17
Added short delay (250ms) between target-disconnect and gdb-exit

EvilEli:
This at (least for me) solves #629 
However I am by no means a TypeScript developer and have no idea if the delay should be implemented like this.
Feel free to improve @haneefdm

### DIFF
--- a/src/backend/mi2/mi2.ts
+++ b/src/backend/mi2/mi2.ts
@@ -393,6 +393,7 @@ export class MI2 extends EventEmitter implements IBackend {
             // program is in paused state
             try {
                 await this.sendCommand('target-disconnect');
+                await new Promise(f => setTimeout(f, 250));
                 this.sendRaw('-gdb-exit');
             }
             catch (e) {


src/backend/mi2/mi2.ts
```diff
@@ -393,6 +393,7 @@ export class MI2 extends EventEmitter implements IBackend {
             // program is in paused state
             try {
                 await this.sendCommand('target-disconnect');
+                await new Promise(f => setTimeout(f, 250));
```
haneefdm, line 396: I don't see how or why this works.  I ask because target-disconnect is failing, causing an exception and going to line 400 (catch block). So, the timeout has no effect -- didn't even get executed.

See [comment](https://github.com/Marus/cortex-debug/issues/629#issuecomment-1128152613)

haneefdm: Yeah, that is a good idea looking back. But now, it is too late.

This is so frustrating. the end of the session is rife with landmines. Things fail at every possible point. And for not explainable reason. I can accept your solution as is (with some analysis) and update #629

Now, there are two+ timeouts in this code path. My worry is that they conflict.

EvilEli: It would have been nice if the issues were due to the same origin.

I totally can feel that frustration. I hope there are still hints coming up on that topic.

Nevertheless thanks for analysing the PR.

haneefdm: Just to be sure, which gdb-server are you using? I cannot get the timeout to do anything for me with JLink.

haneefdm:
Hi @EvilEli Could you try the following replacement for stop?
```typescript
    public async stop() {
        if (trace) {
            this.log('stderr', 'stop');
        }
        if (!this.exited) {
            // With JLink all of these catches, timeouts occur one time or the other. Two back to back runs don't produce
            // the same program flow. Sometimes, we get all the way to a proper gdb-exit without any timers expiring and
            // everything working. Very next run totally erratic
            let timer;
            const startKillTimeout = (ms: number) => {
                if (timer) { clearTimeout(timer); }
                timer = setTimeout(() => {
                    if (timer && !this.exited) {
                        ServerConsoleLog('GDB Kill timer expired for a disconnect+exit, so forcing a kill', this.pid);
                        this.tryKill();
                    }
                    timer = undefined;
                }, ms);
            };
            const destroyTimer = () => {
                if (timer) {
                    clearTimeout(timer);
                    timer = undefined;
                }
            }
            this.process.on('exit', (code) => { destroyTimer(); });
            // Disconnect first. Not doing so and exiting will cause an unwanted detach if the
            // program is in paused state
            try {
                startKillTimeout(500);
                await new Promise((res) => setTimeout(res, 100));       // For some people delay was needed. Doesn't hurt I guess
                await this.sendCommand('target-disconnect');            // Yes, this can fail
            }
            catch (e) {
                if (this.exited) {
                    ServerConsoleLog('GDB already exited during a target-disconnect', this.pid);
                    destroyTimer();
                    return;
                }
                ServerConsoleLog(`target-disconnect failed with exception: ${e}. Proceeding to gdb-exit` + e, this.pid);
            }

            startKillTimeout(350);                                  // Reset timer for a smaller timeout
            await new Promise((res) => setTimeout(res, 250));       // For some people delay was needed. Doesn't hurt I guess
            if (this.exited) {
                // This occurs sometimes after a successful disconnect.
                ServerConsoleLog('gdb already exited before an exit was requested', this.pid);
                return;
            }
            this.sendRaw('-gdb-exit');
        }
    }
```

haneefdm:
@EvilEli I just pushed my changes (including yours) to master. Could you try with that to see if your own issues are resolved?

Unfortunately, too many other unrelated changes. Hopefully, they don't get in the way.

haneefdm:
Just to get feedback, I reopened

EvilEli:
> Just to get feedback, I reopened.
> 
> And, no hurry, don't mean to pressure you
@haneefdm 
I tried your changes and can confirm that this is working for me.
Thanks a lot for your support.

PR #664, opened 2022-06-02
Feature/rtos view ucos ii merge fix

haneefdm:


PhilippHaefele:
Thanks and you where very fast on this. 😄

PhilippHaefele:
You could have reset the master before we merge to remove the mess, but it's also ok in this way.

For others. This is a follow up merge request of #663

PR #673, opened 2022-06-09
backend/disasm.ts: fix typo: instrunctions -> instructions

alefminus:
Tiny typo fix. Looks great so far, I'm using it with an XDS110 instead of code composer studio, which I find buggy on linux (fedora 36). Thanks for all the hard work!


haneefdm:
Thanks.

PR #694, opened 2022-07-12
Return memory references from evaluate requests

benmcmorran:
This functionality is needed to support https://github.com/microsoft/vscode-embedded-tools/issues/3.

Per the DAP spec Embedded Tools treats `memoryReference` values as opaque. This means in order to issue a `readMemory` request, we must first get a memory reference from an `evaluate` or `variables` request. However, Cortex-Debug does not currently return a `memoryReference` from an `evaluate` request, making it difficult or impossible to read some values from memory without directly constructing the `memoryReference`.

@robotdad @gcampbell FYI


haneefdm:
Thank you Ben

PR #700, opened 2022-07-12
Support channels 1-15 for JLink RTT

DepthDeluxe:
* JLinkGDBServer supports setting the RTT channel number via control
  commands set within the first 100ms of established TCP connection.

* JLinkGDBServer does not allow multiple concurrent TCP connections to
  the same TCP RTT port so we are still blocked from running multiple
  RTT channels concurrently.

I validated these changes by attaching a debug session to an RTT-enabled MCU that outputs logs on RTT channel 2.  Before these changes, input validation instructed me to change RTT channel to 0.  After removing input validation, the RTT terminal didn't show up.  After putting in the full RTT configuration implementation, application logs started streaming to the debug terminal as expected.

Tagging @haneefdm since it looks like they most recently contributed to RTT-related work.

### DIFF
--- a/src/frontend/configprovider.ts
+++ b/src/frontend/configprovider.ts
@@ -420,12 +420,18 @@ export class CortexDebugConfigurationProvider implements vscode.DebugConfigurati
         }
 
         if (config.rttConfig && config.rttConfig.enabled && config.rttConfig.decoders && (config.rttConfig.decoders.length !== 0)) {
+            var chosenPort = undefined;
             for (const dec of config.rttConfig.decoders) {
                 if (dec.port === undefined) {
                     dec.port = 0;
-                } else if (dec.port !== 0) {
-                    return `Invalid port/channel '${dec.port}'. JLink RTT port/channel has to be 0 because that is the only one supported by JLinkGDBServer`;
+                } else if (dec.port < 0 || dec.port > 15) {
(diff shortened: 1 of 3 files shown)


haneefdm:
So, unlike OpenOCD, we are still limited to ONE RTT channel but it can be anything between 0-15? Before, it implicitly checked to make sure that there is only one channel and that channel better be zero. I see that check missing now (I think).

haneefdm:
Note that we allowed multiple decoders on a given channel. In JLink's case, there just has to be one as the gdb-server does not support more than one.

I know there are other utilities that can also support another channel but not with the gdb-server.

haneefdm:
Forgot to be explicit. Can you make the change to make sure only one channel is used regardless of what that is.

DepthDeluxe:
@haneefdm, your first comment is correct.  You can only use a single RTT channel when debugging but with these changes, the RTT channel can be 0-15 instead of just 0. I can add the check to ensure there is only one RTT channel used.

DepthDeluxe:
@haneefdm, fixed.  I confirmed that I could attach 2 decoders, 1 binary and 1 console, to the same non-zero RTT channel and everything works as expected.  I also confirmed that if I try to use 2 decoders with 2 different channels, VSCode displays the error message added to this PR.

haneefdm:
Thanks. Merged

PR #705, opened 2022-07-18
embOS RTOS view support

PhilippHaefele:
"Blocked" by https://github.com/Marus/cortex-debug/pull/704

Adds support for embOS

Tested with Segger  example from https://www.segger.com/downloads/embos/ => embOS for Cortex-M and GCC V5.16.1.0 => STM32F429 (yes finally got one 😄)

Did add some code to handle older versions of embOS but i didn't test them due to lack of an older example version.

### DIFF
--- a/src/frontend/rtos/rtos-embos.ts
+++ b/src/frontend/rtos/rtos-embos.ts
@@ -38,11 +38,11 @@ export class RTOSEmbOS extends RTOSCommon.RTOSBase {
     // We keep a bunch of variable references (essentially pointers) that we can use to query for values
     // Since all of them are global variable, we only need to create them once per session. These are
     // similar to Watch/Hover variables
-    private OS_Global: RTOSCommon.RTOSVarHelper;
-    private OS_GlobalVal: any;
+    private OSGlobal: RTOSCommon.RTOSVarHelper;
+    private OSGlobalVal: any;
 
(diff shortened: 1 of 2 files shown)


PhilippHaefele:
Thanks 😃

haneefdm:
I will update the change log.

Sorry, I have been super busy and then super (the ICU variety) sick. I **NEED** to make a release **SOON**. Piling up.

News: Microsoft now supports Cortex-Debug as a debugger they recognize -- IAR is the other one. So, their RTOS/embedded tools will also work with our extension. They even made a PR to make things work for them. I had a short meet and greet over the phone with them. That was nice of them.

So, now will have Azure as well, albeit a different extension.

PhilippHaefele:
> I will update the change log.

Thanks 😃

> Sorry, I have been super busy and then super (the ICU variety) sick. I NEED to make a release SOON. Piling up.

Too bad to hear that. Get well soon.

No deal at least my team can wait for it so far. Health is more important than that.

> News: Microsoft now supports Cortex-Debug as a debugger they recognize -- IAR is the other one. So, their RTOS/embedded tools will also work with our extension. They even made a PR to make things work for them. I had a short meet and greet over the phone with them. That was nice of them.

Cool news and nice they contacted you via phone for that 👍 

> So, now will have Azure as well, albeit a different extension.

The cross usage of debug extensions seems little bit odd for me, but as it works why not

PR #718, opened 2022-08-18
Add External JLink servertype.

royb3:
I am using a setup where I have a JLink GDB server running on an external system a lot.

The external server type worked for this, but it had errors on unexpected arguments for `monitor reset`

I coppied the external server type and replaced the launch/attach/restart commands to match the ones from the JLink server type.

Do you think this is the right approach or would you know a better one?

### DIFF
--- a/package.json
+++ b/package.json
@@ -564,7 +564,7 @@
                         "properties": {
                             "servertype": {
                                 "type": "string",
-                                "description": "GDB Server type - supported types are jlink, openocd, pyocd, pe, stlink, stutil, qemu, bmp and external",
+                                "description": "GDB Server type - supported types are jlink, openocd, pyocd, pe, stlink, stutil, qemu, bmp, jlinkexternal and external",
                                 "enum": [
                                     "jlink",
                                     "openocd",
@@ -574,7 +574,8 @@
                                     "bmp",
                                     "pe",
                                     "qemu",
-                                    "external"
+                                    "external",
+                                    "jlinkexternal"
                                 ]
                             },
                             "cwd": {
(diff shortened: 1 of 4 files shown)


haneefdm:
It is not clear to me what is different. With the `external` server modeled after OpenOCD has generic stuff. You are supposed to customize certain things. It will never work for all the gdb-servers out there.

Could you present a diff between JLink (or external) and your new server type? I would have thought that you would derive from the JLink server itself and not duplicate all the code.

Basically, I want to see what is actually different. Also, I would have thought you would use the remote access method rather than this way. If it was just the reset mechanism, JLink always had reset mechanisms that are not properly documented. For instance, there is a number after the reset and for a given vendor it seems mean something different. For this purpose, we always had the overrideXXX settings in launch.json. Not just for JLink but other servers as well. Some devices have super complicated or different reset mechanisms.

https://www.segger.com/products/debug-probes/j-link/tools/j-link-remote-server.
https://mcuoneclipse.com/2017/02/05/remote-board-debugging-j-link-remote-server-with-eclipse/comment-page-1/
https://wiki.segger.com/J-Link_GDB_Server#-select

See the -select option in the last link. If you want to work with me, we should try enabling the remote access so that all things like SWO/RTT/etc. can also work over a tunnel. For whatever reason you are using the external option, this may be a better way. There are many people already using this and it used by people during manufacturing or running CI/CD tasks.

royb3:
I had not found out about the overrideLaunchCommands / overrideRestartCommands yet. That should be sufficient for my usecase.

haneefdm:
https://github.com/Marus/cortex-debug/blob/master/debug_attributes.md

dsabala:
> If you want to work with me, we should try enabling the remote access so that all things like SWO/RTT/etc. can also work over a tunnel. 
@haneefdm So now RTT is not supported when server type is external?
```
      "gdbTarget": "localhost:50000",
      "servertype": "external",
```

dsabala:
Actually it is not a problem, I managed to run JLink suite inside container and bind USB devices
```
{
    "name": "my-devcontainer",
    "image": "my-container:latest",
    "runArgs": [
        "--privileged",
        "-v", "/dev/bus/usb:/dev/bus/usb"
    ],
    "customizations": {
        "vscode": {
          "extensions":
            [
                "mcu-debug.debug-tracker-vscode",
                "ms-vscode.cpptools",
                "marus25.cortex-debug"
            ]
        }
    }
}
```
Everything works great (building, debugging, RTT) and it offers non-containerized experience ;)

haneefdm:
Thanks for reporting back

PR #735, opened 2022-09-01
Add BMP connect_srst option

andrej-mk:
Adds option `bmpConnectUnderReset` support to launch configuration.
When true,  `monitor connect_srst` is executed before swdp scan. Probe then issues reset to target before connecting.

https://black-magic.org/usage/gdb-commands.html


haneefdm:
Thanks. But looks like TSLint failed. VSCode should have reminded you as it is part of the workspace settings....

While TSLint is obsolete, we have not moved to ESLint -- super large effort, so we still rely on TSLint

andrej-mk:
Sorry, I missed a semicolon and I didn't have lint extension installed. I've fixed my commit.

PR #742, opened 2022-09-14
Incorrect naming for srst

vikkasSWE:
No command `monitor connect_rst disable` availible in gdb. 
Correct is `monitor connect_srst disable`.

Discovered from 2h debugging trying to flash an STM32F1 with a black magic probe and 0 search result on google.

### DIFF
--- a/src/bmp.ts
+++ b/src/bmp.ts
@@ -32,9 +32,9 @@ export class BMPServerController extends EventEmitter implements GDBServerContro
         ];
 
         if (this.args.bmpConnectUnderReset) {
-            commands.push('interpreter-exec console "monitor connect_rst enable"');
+            commands.push('interpreter-exec console "monitor connect_srst enable"');
         } else {
-            commands.push('interpreter-exec console "monitor connect_rst disable"');
+            commands.push('interpreter-exec console "monitor connect_srst disable"');
         }
 
         if (this.args.powerOverBMP === 'enable') {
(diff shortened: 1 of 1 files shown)


haneefdm:
We don't maintain BMP, I don't even own one. This is currently done by user contributions.

Those lines were added by @andrej-mk on Sept 1. Not sure how it was tested and why it is wrong.

https://github.com/Marus/cortex-debug/pull/735

But clearly, it is wrong according to BMPs documentation.

andrej-mk:
I tested hosted BMP connected to stlink. connect_srst doesn't work there, connect_rst does.
bug

haneefdm:
@andrej-mk Can you do a `monitor help` on the gdb prompt?

Can we also compare versions of BMP? Perhaps one of you is not recent. But if we look at the documentation

https://black-magic.org/usage/gdb-commands.html

`connect_rst` is not a valid command. Maybe this is device dependent. Some boards don't have 'SRST`. Perhaps you can contact BMP folks to see what the deal is.

We need to resolve this between @vikkasSWE and @andrej-mk or I am inclined to back out the change on Sept 1st.

andrej-mk:
I was using BMP master, as I was doing some changes there. [Commit](https://github.com/blackmagic-debug/blackmagic/commit/94afb264e4fa188dad5702c39d8b1eb9b49318ae) I posted changed the option, but it isn't released yet. For current BMD code, `connect_srst` is valid.

haneefdm:
Hmmm. This is not sustainable if BMP keeps changing like that. I see that when bmp is released, it will break again if we move to the current srst. Perhaps this change was not needed at all. This is why we have `overrideLaunchCommands`

haneefdm:
Also, the documentation does not say what the default is. If no enable/disable is issued, what is the default?

andrej-mk:
It is disabled by default. If it is possible to use `overrideLaunchCommands` to enable this option, I'm fine with reverting my commit.

PR #743, opened 2022-09-14
Revert "Add BMP connect_srst option (#735)"

andrej-mk:
This reverts commit d4858f0b9256b5f1df7cae4a2c29975b559680a1.

Same result can be acheived with overrideLaunchCommands.


haneefdm:
Thanks.

PR #753, opened 2022-10-06
Add preContinueCommands configuration option

wayneuroda:
I find it is very convenient to have some kind of global flag in my firmware which can disable certain features, such as watchdog and other system health monitors during debug (I know many processors halt the watchdog during break, however a software monitor based on tick count or time etc may not be paused).

I found no way to set the flag and then re-set it through device resets in the debugger. I could not use the `postRestartCommands` as these happen too early in the boot process (thus the flag is zeroed out by the crt initialization, which is what we do want!) and/or the target is not in break mode at the time when `postRestartCommands` are run.

Perhaps there is a way to execute a command eg `set debugger_attached=1` both on attach and also after the "Reset device" button is pressed, but I couldn't figure it out ;)

Using this new configuration as well as `runToEntryPoint`, after pressing "Reset device", the debugger will break at eg. main, and then when I press the continue button it will set the debugging flag as I want!

### DIFF
--- a/debug_attributes.md
+++ b/debug_attributes.md
@@ -30,6 +30,7 @@ Also using IntelliSense while editing launch.json in VSCode can be quite helpful
 | preAttachCommands | Common | Additional GDB Commands to be executed at the start of the main attach sequence (immediately after attaching to target).
 | preLaunchCommands | Common | Additional GDB Commands to be executed at the start of the main launch sequence (immediately after attaching to target).
 | preRestartCommands | Common | Additional GDB Commands to be executed at the beginning of the restart sequence (after interrupting execution).
+| preContinueCommands | Common | Additional GDB Commands to be exectuted immediately before continuing.
 | request | Common | ????
 | rtos | Common | RTOS being used. For JLink this can be Azure, ChibiOS, embOS, FreeRTOS, NuttX, Zephyr or the path to a custom JLink RTOS Plugin library. For OpenOCD this can be ChibiOS, eCos, embKernel, FreeRTOS, mqx, nuttx, ThreadX, uCOS-III, or auto.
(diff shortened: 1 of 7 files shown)


haneefdm:
I don't want to accept this PR. This is very specific and somehow tied to a continue action. Seems to me, what you really want is a hook on a stop where you can even examine your program and do what is needed

All of these can be done in gdb itself without changing this debugger and it will be a  portable flow.

https://undo.io/resources/gdb-watchpoint/how-use-gdb-command-hook/
https://sourceware.org/gdb/onlinedocs/gdb/Hooks.html
https://sourceware.org/gdb/onlinedocs/gdb/Define.html

Note that you do not get all the actions mentioned in the hooks page because we do not use those commands. We use the machine interface. You can have the following in your .gdbinit in the current dir
```
define hook-stop
echo hook-stop\n
end
```

And you will see that the hook is called every time program stops for whatever reason. gdb has an entire scripting language if if-then-elses, loops, etc. You can access (read/write) registers, variables, etc. You can probably even undefined your hook when no longer needed.

You can also use the python embedded version of gdb to do even more magic.

flaik-wayne:
Yes, you are right. And I honestly did try to use `hook-stop` previously... now it works. I think what confused me before is that I thought `hook-stop` was not being run because the contents of it are not printed inside the Debug Console, but of course, it has nothing to do with the `mi` connection so why would it show up in the console there. It does work. I will close the PR.

wayneuroda:
Yes, you are right. And I honestly did try to use hook-stop previously... now it works. I think what confused me before is that I thought hook-stop was not being run because the contents of it are not printed inside the Debug Console, but of course, it has nothing to do with the mi connection so why would it show up in the console there. It does work. I will close the PR.

PR #758, opened 2022-10-17
Fix handling of optional parameters in stackTraceRequest

jedrzejboczar:
Hi, I was trying to use cortex-debug with [nvim-dap](https://github.com/mfussenegger/nvim-dap) and noticed an issue with stackTrace reqest. According to [DAP specification](https://microsoft.github.io/debug-adapter-protocol/specification#Requests_StackTrace), `startFrame` and `levels` are optional and if ommitted it means "start from 0, all frames", but cortex-debug will use `undefined` in calculations and the resulting command contains literal `undefined` instead of numerical values.

### DIFF
--- a/src/gdb.ts
+++ b/src/gdb.ts
@@ -2309,6 +2309,9 @@ export class GDBDebugSession extends LoggingDebugSession {
             stackFrames: [],
             totalFrames: 0
         };
+        // Handle optional args when not passed
+        args.startFrame = args.startFrame ?? 0;
+        args.levels = args.levels ?? Infinity;
         if (!this.isMIStatusStopped() || !this.stopped || this.disableSendStoppedEvents || this.continuing) {
             this.sendResponse(response);
             return Promise.resolve();


haneefdm:
On the surface, things look okay. We probably had that since 4-5 years and ago and the DAP is moved on which we found impossible to keep track of. If you started somewhere in the timeline, and we did much earlier, you find stuff like this.

Thanks for the PR.

Let me check the DAP again and our implementation will approve the PR. Technically, it should be equivalent even without the PR but I see that we are using a value in calculations.

haneefdm (commented):
I don't have the complete history of DAP or this function. The `args.startFrame` change makes sense but VSCode as a client always supplied a value. Since neovim is a new client, you may find many such issues in our and others' extensions.

jedrzejboczar:
Thanks for quick response. I understand that it's problematic to keep track of DAP specification and that this extension is mostly targeted at VS Code, but I just wanted to try if it would work with neovim. I actually managed to get it working with very little changes. I noticed that stackTrace does not handle these optional args because nvim-dap sends only threadId, so I created this PR, as I think this change shouldn't cause any problems with VS Code compatibility.

haneefdm:
There are a lot of unsaid things about clients of the DAP. VSCode is the first client everyone tries to make happy but they also make some smart decisions.

How nvim is using this DAP is not right. It is guaranteed to cause hangs...trust me on this. VSCode

DAP on the other hand tries to be general and does not specify how error situations are to be taken care of. In this enable DAP says
```typescipt
  /**
   * The maximum number of frames to return. If levels is not specified or 0,
   * all frames are returned.
   */
  levels?: number;
```
**ALL FRAME ARE RETURNED**. No, that option should never be used by a client. Always have a small/reasonable limit, because you can always ask for more

There are dozens of such things in the DAP that are not practical. We all find out eventually :-)

haneefdm:
Btw, we protect ourselves from a super deep stack by limiting the maxDepth to 1000 or else gdb hangs. Is 1000 the right limit? It is what it is. Something is wrong if your program is doing that.

jedrzejboczar:
I guess that makes sense, 1000 frames sounds like more than enough. I think I'll create a PR to nvim-dap instead to just use some fixed value for level.

haneefdm:
Just to be clear, here is what VSCode as a client does

* it only asks for 20 frames at a time -- about how much the window can fit with the UI capable to asking for more
* it keeps asking for 20 more so everything is responsive.

We as the DAP provider, however, keep doling out whatever frames are asked for until we get to a 1000 and we let VSCode (the client) know, there are no more. VSCode does not itself (AFAIK) enforce a limit

It is all about being performant and responsive and just asking for 20 frames seems to do the job. Handling the 99.9% use case.

jedrzejboczar:
I'll try implementing something similar in nvim-dap, as it seems like a reasonable approach. Thanks for advice! I'll close this PR now.

haneefdm:
No, I will merge it anyways and change Infinity to max safe integer. Very reasonable to do.

jedrzejboczar:
Ok, thanks. I will create a PR limiting in nvim-dap as you suggested.

PR #761, opened 2022-10-21
pre,post and override Reset commands

ciscoski:
* Reset and Restart command will now behave differently
* preRestartCommands, postRestartCommands, overrideRestartCommands will work only for restart.
* preResetCommands, postResetCommands, overrideResetCommands will work only for reset.
* By default Reset and Restart default comands are identical.

This allows the user to not have to stop/restart the debug session if a new build is needed.
Example before PR
```json
{
    "version": "0.2.0",
    "configurations": [
        {
            "name": "Cortex Debug - PyOCD",
            "cwd": "${workspaceFolder}",
            "request": "launch",
            "type": "cortex-debug",
            "runToEntryPoint": "main",
            "servertype": "pyocd",
            "targetId" : "stm32f217ietx",
            "serverArgs": [
                "--pack=C:/Users/Cisco.Cervellera/Downloads/Keil.STM32F2xx_DFP.2.9.0.pack",
                "--semihosting",
                "-Osemihost_console_type=console",
            ],
            "svdFile": "C:/Users/cisco.cervellera/Downloads/STM32F21x.svd",
            "showDevDebugOutput": "raw",
            "preLaunchTask" : "${defaultBuildTask}",
            "preRestartCommands": [
                "load"],
        },
    ]
}
```
This would call load everytime also on reset (which is not needed and time consuming).
With the proposed PR the `load` comand will be called only on Restart which will also call the `defaultBuildTask`.

The Wiki will needed to be updated to reflect this changes.

Happy to answare question.

### DIFF
--- a/debug_attributes.md
+++ b/debug_attributes.md
@@ -22,14 +22,17 @@ Also using IntelliSense while editing launch.json in VSCode can be quite helpful
 | overrideGDBServerStartedRegex | Common | You can supply a regular expression (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions) in the configuration property to override the output from the GDB Server that is looked for to determine if the GDB Server has started. Under most circumstances this will not be necessary - but could be needed as a result of a change in the output of a GDB Server making it incompatible with cortex-debug. This property has no effect for bmp or external GDB Server types.
 | overrideLaunchCommands | Common | You can use this to property to override the commands that are normally executed as part of flashing and launching the target. In most cases it is preferable to use preLaunchCommands and postLaunchCommands to customize the GDB launch sequence.
 | overrideRestartCommands | Common | You can use this to property to override the commands that are normally executed as part of restarting the target. In most cases it is preferable to use preRestartCommands and postRestartCommands to customize the GDB restart sequence.
+| overrideResetCommands | Common | You can use this to property to override the commands that are normally executed as part of reset the target. When not defined this will have the same value of overrideRestartCommands. In most cases it is preferable to use preResetCommands and postResetCommands to customize the GDB reset sequence.
 | postAttachCommands | Common | Additional GDB Commands to be executed after the main attach sequence has finished.
 | postLaunchCommands | Common | Additional GDB Commands to be executed after the main launch sequence has finished.
 | postRestartCommands | Common | Additional GDB Commands to be executed at the end of the restart sequence.
+| postResetCommands | Common | Additional GDB Commands to be executed at the end of the reset sequence. When not defined this will have the same value of postRestartCommands.
 | postRestartSessionCommands | Common | Additional GDB Commands to be executed at the end of the re-start sequence, after a debug session has already started.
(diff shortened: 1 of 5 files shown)


haneefdm:
Thanks for the PR.

> Happy to answare question.

#1 requirement is that it has to be 100% compatible. Has this been met? I am not reviewing this PR until that is addressed. Let me know.

ciscoski:
Can you please specify what 100% compatible means ?
I could not find policies for contribution.

haneefdm:
This means if I make zero changes to my launch.json, everything works exactly the same with Reset and Restart

ciscoski:
This is not a 100% compatible change.

I would argue that this PR is almost fixing an issue that reset and restart are treated the same way from the backend.
Before the change `preRestartCommands` and `postRestartCommands` would work for both reset and restart which (imho)  is logically not correct.
I understand that reset does not invoke VSCode protocol (rebuild) and as such should have its own pre post commands.

I could make this 100% compatible but it would make `preResetCommands`, `postResetCommands` and `overrideResetCommands`  attibutes convoluted and not easy to understand.

haneefdm:
Compatibility is non-negotiable. It is not that hard to make it compatible.

* You have two sets of commands - one for reset (new ones) and one for restart (existing ones). Lets agree on that first
* For `Reset`, we will defer to the `Restart` version of the settings if a `Reset` specific one does not exist

I can absolutely guarantee you thousands of users' launch.json will break. I know that for a fact.

> I would argue that this PR is almost fixing an issue that reset and restart are treated the same way from the backend.

Nice try :-) There was a time when VSCode changed its mind monthly -- on what a Restart did. We created a Reset so we would have guaranteed behavior as VSCode kept flip-flopping. Granted reset should have had its own set of settings.  So, that is water under the bridge and not a reason to break compatibility.

When you are done with this, I would like to add a feature to `reloadProgramOnRestart` which will do the reload properly or even do a true Restart. Reloading programs is much more complicated in some cases -- especially multi-core devices. Maybe this is a better option than having reset-specific options and that is what would be configurable. No need for two sets.

so, please take compatibility super seriously. You are already working in code that is very critical and complex and there are so many use cases that can break (runToEntryPoint, breakOnReset, etc.).

haneefdm:
This is one of the reasons, I want people to start with an Issue, and propose/debate solutions working towards a PR. Not start with a PR.

ciscoski:
> Nice try :-) There was a time when VSCode changed its mind monthly -- on what a Restart did. We created a Reset so we would have guaranteed behavior as VSCode kept flip-flopping. Granted reset should have had its own set of settings. So, that is water under the bridge and not a reason to break compatibility.

Opinable but understandable. 

> When you are done with this, I would like to add a feature to reloadProgramOnRestart which will do the reload properly or even do a true Restart. Reloading programs is much more complicated in some cases -- especially multi-core devices. Maybe this is a better option than having reset-specific options and that is what would be configurable. No need for two sets.

If I've time/need I can have a look :).

I've changed the PR to accomodate the 100% compatibility and linting problem.

PR #813, opened 2023-01-31
Add peripheral tree refresh and collapse commands

### LINKED ISSUE #803: In SVD viewer, the "Refresh" button updates the registers of only one peripheral in which the cursor is located
Снимок экрана 2023-01-17 104837

thegecko:
As outlined in #803, the tree refresh command only refreshes the highlighted node. This is due to the command passing the selected node to the `peripheralsForceRefresh()` function.

This PR adds a new command for refresh which ignores the selected node. The functionality was also missing to recurse the trees and refresh the expanded nodes, so this has also been added.

As suggested, I've also exposed a `Collapse All` command.

Fixes #803

### DIFF
--- a/package.json
+++ b/package.json
@@ -447,6 +447,16 @@
                 "icon": "$(refresh)"
             },
             {
+                "command": "cortex-debug.peripherals.refreshAll",
+                "title": "Refresh All",
+                "icon": "$(refresh)"
+            },
+            {
+                "command": "cortex-debug.peripherals.collapseAll",
+                "title": "Collapse All",
+                "icon": "$(collapse-all)"
+            },
+            {
                 "command": "cortex-debug.peripherals.pin",
                 "title": "Pin",
                 "icon": "$(pin)"
@@ -2869,6 +2879,14 @@
                     "when": "false"
                 },
                 {
+                    "command": "cortex-debug.peripherals.refreshAll",
+                    "when": "false"
+                },
+                {
+                    "command": "cortex-debug.peripherals.collapseAll",
+                    "when": "false"
+                },
+                {
                     "command": "cortex-debug.peripherals.pin",
                     "when": "false"
                 },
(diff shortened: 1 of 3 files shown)


haneefdm:
We already have a method to update the entire tree for a session and do it only for 

PeripheralTreeForSession.updateData()

Bigger issue is we should only call this for sessions that are in a stopped state. Otherwise, we will create a lot of errors as we fire a bunch reads almost in parallel.  Right now, we do not keep the stopped state in the tree and that has to be added and is trivial to add. Could you do that please?

If you iterate over the top level set of trees, then just calling updateData on each tree whose session is in stopped state will do the trick.

We must always remember that we are designed for multi-session mode.

thegecko:
> We already have a method to update the entire tree for a session and do it only for

So what session should be refreshed if many are running? Just the active one?

> PeripheralTreeForSession.updateData()

this is called https://github.com/Marus/cortex-debug/pull/813/files#diff-46c4c48f00f5d767470da6cf4f9bc1ee04aeb6256897e829fcac63f794f889dfR221

We just need to decide if this is for all sessions or the `vscode.debug.activeDebugSession`.

I can add a guard to check sessions are stopped.

haneefdm:
Refresh all stopped ones. 
https://github.com/Marus/cortex-debug/blob/056c03f01e008828e6527c571ef5c9adaf64083f/src/frontend/views/peripheral.ts#L261

We can set and clear a flag on each tree when Stopped and Continued. Then, you can query that flag while implementing a refreshAll()

```
    public debugStopped(session: vscode.DebugSession) {
        const regs = this.sessionPeripheralsMap.get(session.id);
        if (regs) {     // We are called even before the session has started, as part of reset
            regs.updateData();
        }
    }

    public debugContinued() {
    }
```

haneefdm:
Oh and `debugContinued()` will need a session argument like `debugStopped()` We should also do something in sessionTerminated() to clear the stopped flag. I will address that and a couple of other things later. All the peripheral node-related commands/events haven't been touched in years and it was before I started working on this repo.

haneefdm:
Looks like CI failed on lint

ERROR: /home/runner/work/cortex-debug/cortex-debug/src/frontend/views/peripheral.ts:237:40 - ` should be '

thegecko:
> Looks like CI failed on lint

Fixed. I wasn't aware the CI ran lint, perhaps the lint command should be exposed in a package.json script?

haneefdm:
I thought TSLint is run automatically when I Open the folder in VSCode

One issue is TSList has been deprecated but we still use it and you have to install it forcefully. Other than that, it should be part of the build process. I never run it separately.

One of these days, we will more to new versions of TS and migrate to ESLint. Tried twice and botched it. Now that there is less and less source code, I will try again sometime.

haneefdm:
Yes, all the stuff under mcu-debug uses ESLint.

PR #825, opened 2023-02-10
Feature/dl.pemicro swo

dlizewski:
SWO support using peMicro debugger streaming interface.

Mostly reverse engineered from https://www.pemicro.com/blog/index.cfm?post_id=165

Supported Hardware:
Multilink ACP Rev. B and later
Multilink Universal Rev. D and later
Multilink Universal FX Rev C and later

I only have access to a Multilink Universal Rev. D
The other hardware versions are untested.

I made it as an extension to the regular socket class. Let me know if you think there's a better place for it.

I also fixed a bug in the ITM decoder overwriting the buffer if a timestamp is malformed and doesnt complete in 5 bytes.
It was causing an infinite loop of exceptions for peMicro after a clock change.

### DIFF
--- a/src/frontend/swo/core.ts
+++ b/src/frontend/swo/core.ts
@@ -69,8 +69,10 @@ class ITMDecoder extends EventEmitter {
     }
 
     private rxWriteByte(byte: number): boolean {
-        this.rxBuffer.writeUInt8(byte, this.rxCount);
-        this.rxCount++;
+        if(this.rxCount < this.rxBuffer.length) {
+            this.rxBuffer.writeUInt8(byte, this.rxCount);
+            this.rxCount++;
+        }
         return this.rxCount === this.rxTargetLength;
     }
 
@@ -135,8 +137,16 @@ class ITMDecoder extends EventEmitter {
                     break;
                 case Status.TIMESTAMP:
                     this.rxWriteByte(byte);
+                    // Check if the continuation bit is false.
+                    // This indicates the last byte in a timestamp
                     if ((byte & 0x80) === 0x00) {
                         this.emit('timestamp', this.getRxPacket());
+                        newStatus = Status.IDLE;
+                    } else if (this.rxCount == 5) {
+                        // A timestamp is at most 5 packets. If we didn't see the continuation bit false, something has gone wrong
+                        // This often happens with PeMicro because it starts sending garbage after a clock change.
+                        // In theory we should go to UNSYNCED, but that leads to never recovering.
+                        // Going back to IDLE allows recovery when it stops sending garbage
                         newStatus = Status.IDLE;
                     }
                     break;
--- a/src/frontend/swo/core.ts
+++ b/src/frontend/swo/core.ts
@@ -69,7 +69,7 @@ class ITMDecoder extends EventEmitter {
     }
 
     private rxWriteByte(byte: number): boolean {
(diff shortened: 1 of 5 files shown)


src/frontend/swo/core.ts
```diff
@@ -69,8 +69,10 @@ class ITMDecoder extends EventEmitter {
     }
 
     private rxWriteByte(byte: number): boolean {
-        this.rxBuffer.writeUInt8(byte, this.rxCount);
-        this.rxCount++;
+        if(this.rxCount < this.rxBuffer.length) {
+            this.rxBuffer.writeUInt8(byte, this.rxCount);
+            this.rxCount++;
+        }
```
haneefdm, line 75: Could you explain what is going on here? Please refresh my memory about the protocol. Maybe you fixed a problem that already existed but I can't tell. If you explain, it will save me a bunch of time researching stuff

dlizewski: The main issue was below in the case Status.TIMESTAMP

This technically isn't an issue for a well formed ITM packet stream but I found PeMicro doesn't always produce a  well formed stream.

A "local timestamp" packet is a variable length packet with 1-5 bytes. The end of timestamp is marked by the upper most bit "byte & 0x80".

The issue with the code is that if the stream initiated a Status.TIMESTAMP and the end bit "byte & 0x80" didnt come in the first five bytes, then on the 6th byte this.rxWriteByte(byte); would overflow this.rxBuffer raising an uncaught exception. (this.rxBuffer is length 5 because timestamp is max 5 bytes)

The problem, at least in the case of PeMicro is that during a clock change the stream could start temporarily send garbage. This state machine would get stuck in case Status.TIMESTAMP because it kept overflowing this.rxBuffer even after the stream started sending good data again (because it wasn't timestamp data anymore).

I made it so it counted the number of received bytes in "case Status.TIMESTAMP" (by checking the return of rxWriteByte) and will exit that state if it received more than 5 since it couldnt be a valid timestamp anymore.

https://developer.arm.com/documentation/ddi0403/d/Appendices/Debug-ITM-and-DWT-Packet-Protocol/Packet-descriptions/Local-timestamp-packets?lang=en

Technically I could remove the change in rxWriteByte and let it throw an exception since the change in case Status.TIMESTAMP prevents the overflow as well.

dlizewski: It might be a good idea to keep rxWriteByte as it was and only include the change in "case Status.TIMESTAMP". Preventing the overflow in rxWriteByte might mask future problems

haneefdm:
I was about to do a merge...

Sorry, lint failed. You have to install and enable TSLint. VSCode deperacated TSLint but we still have to use it until we migrate everything to ESLint which turns out of ba HUGE task.

dlizewski:
Thanks for looking at my PR. I've fixed the lint issues identified with TSLint, although I cant confirm the CI will pass because it looks like I cant re-run it on the new commits by myself.

Also wanted to express my appreciation for this extension ( and the collection of mcu-debug extensions). I've been using them for a while and they are super useful!

haneefdm:
Yes, the CI was a contribution. It should run on PR creation. I will look into it.

> Also wanted to express my appreciation for this extension ( and the collection of mcu-debug extensions). I've been using them for a while and they are super useful!

Thanks.

PR #844, opened 2023-03-10
Chore/doc cleanup

PhilippHaefele:
@haneefdm Do we still need `Enhanced SVD Auto-selection` in `Planned Features` as the new `Peripheral (SVD) Viewer` now support downloading them from CMSIS Packs?

I do like to add a new screenshot as the many things changed.  
Are there any preferred Color Theme for this?  
I do prefer the `Dark+ (default dark)` one and this one should work for both light and dark themes in the marketplace


haneefdm:
I wasn't even sure what Marcel meant by "Enhanced SVD Auto-selection". But some stuff was added to the svd-viewer repo where it tries to pull in the svd based on deviceName from the Keil CMSIS repository. We can remove that line

As for a screenshot, my preference is the same as yours. `Dark+`

haneefdm:
Thanks. take your time 

Btw, I rewrote the debug_attributes.md generation. It is now done in node and is part of the build. However, `npm watch` will not build it. I am missing one feature of the previous doc. which says which attribute is particular to which gdb-server. I think the document itself should cover that.

PR #851, opened 2023-03-20
Don't assume monitor commands are supported on external servers

MabezDev:
It is not guaranteed that the "monitor" commands will be supported on the external server. This PR removes this assumption. Of course, those who need these commands can still add them manually with the `openOCDLaunchCommands` config.

### DIFF
--- a/src/external.ts
+++ b/src/external.ts
@@ -44,19 +44,14 @@ export class ExternalServerController extends EventEmitter implements GDBServerC
 
     public launchCommands(): string[] {
         const commands = [
-            ...genDownloadCommands(this.args, ['interpreter-exec console "monitor reset halt"']),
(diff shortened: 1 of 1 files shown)


haneefdm:
No, I can't merge this PR. We are going to leave the code as is. Anyone who is using `external` type should be doing override commands (see https://github.com/Marus/cortex-debug/blob/master/debug_attributes.md) in their launch.json to get the right behavior. What the code currently does is provide a template and I do not want to remove that template.

And no, `openOCDLaunchCommands` is not a good way of doing this. It is done via the various override commands.

I will also break it for those who are relying on it (because they are using openocd for instance).

This is more of a documentation issue.

MabezDev:
Ah gotcha, must have missed that. Sorry for the noise.

PR #859, opened 2023-03-30
add new stmcube path for windows

xoviat:
ST seems to have changed the default installation path for STMCubeIDE. Has anyone else experienced this?

### DIFF
--- a/src/stlink.ts
+++ b/src/stlink.ts
@@ -8,7 +8,9 @@ import { EventEmitter } from 'events';
 function get_ST_DIR() {
     switch (os.platform()) {
         case 'win32':
-            return 'C:\\ST';
+            const oldDirName = 'C:\\ST';
+            const newDirName = 'C:\\Program Files\\STMicroelectronics';
+            return fs.existsSync(newDirName) ? newDirName : oldDirName;
         case 'darwin':
             return '/Applications/STM32CubeIDE.app/Contents/Eclipse';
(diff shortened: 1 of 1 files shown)


xoviat:
This appears to work as advertised.

haneefdm:
I am thinking of people who had an old version installed but then installed a newer version.

xoviat:
On second thought, preferring the new installation path is problematic because ST installs other software to that directory also. So if, for example, an old version of STMCubeIDE is installed and a new version of other ST software, these changes would cause cortex-debug to fail. If the old version is preferred, these changes will have no impact because cortex-debug will prefer the old directory first.

haneefdm:
At this point, I have to install ST software myself and see how many ways they are doing it now.

We use very little out of their install. But, being current is the best in our experience. It generates too many support issues when we take old stuff. But I have to deal with that and would like to minimize that. So far, I have not seen a case where the old version of the STLink server or the programmer are needed. But for those cases, we already have a mechanism to force us to use your specific version. What we are talking about here is the default discovery.

So, this PR will stay pending until I know more.

xoviat:
@haneefdm I want to apologize for this PR. ST has not changed the default installation location; I changed the location some months ago myself. Sorry for wasting your time.

haneefdm:
Thanks for letting me know. I am paranoid about breaking people's flow

Issue #755, opened 2022-10-09
defmt rttConfig using JLink


miathedev:
Hi,

how do i add a defmt message decoder to my rttConfig?
When using defmt in combination with rtt decoders the data stream only outputs not readable data, eg:
`�r�Gr'W%2k`

my current launch.json
```
{
    // Use IntelliSense to learn about possible attributes.
    // Hover to view descriptions of existing attributes.
    // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
    "version": "0.2.0",
    "configurations": [
        {
            "name": "Debug Server",
            "request": "launch",
            "type": "cortex-debug",
            "cwd": "${workspaceRoot}",
            "executable": "${workspaceFolder}/target/thumbv7em-none-eabihf/debug/embassy-stm32wl-examples",
            "servertype": "jlink",
            "interface": "swd",
            "device": "STM32WLE5JC",
            "runToMain": true,
            "preLaunchTask": "Cargo build",
            "breakAfterReset": true,
            "showDevDebugOutput": "raw",
            "serialNumber": "XXXXXXXXXXXXX",
            "postLaunchCommands": [
                "help"
            ],            
            "swoConfig": {
                "enabled": true,
                "cpuFrequency": 8000000,
                "swoFrequency": 2000000,
                "source": "probe",
                "decoders": [
                    { "type": "console", "label": "ITM", "port": 0 }
                ],
            },
            "rttConfig": {
                "enabled": true,
                "address": "auto",
                "decoders": [{"port": 0, "type": "console", "noprompt": true, "clear": false, "encoding": "signed"}]
            },
        }
    ]
}
```

tasks.json
```
{
    // See https://go.microsoft.com/fwlink/?LinkId=733558 
    // for the documentation about the tasks.json format
    "version": "2.0.0",
    "tasks": [
        {
            "label": "Cargo build",
            "type": "shell",
            "command": "cargo",
            "args": [
                "flash",
                "--chip",
                "STM32WLE5JCIx",
                "--target",
                "thumbv7em-none-eabihf",
                "--probe",
                "1366:0101:XXXXXXXXXX"
            ],
            "problemMatcher": [
                "$rustc"
            ],
            "group": "build",
        },
    ]
}
```
Because im new to RTT and to be honest Rust in general, i tryed writing over RTT in general without defmt. It worked, but while using **defmt** data is kind of crypted, thats why i guess there must be a decoder for that. Is there one for this plugin? Or is there a workaround?

[...]

PR #880, opened 2023-05-17
Allow printing defmt output to the RTT console

xoviat:
defmt is widely used as a logger in the rust embedded ecosystem. This allows seamlessly debugging a rust program while printing defmt output at the same time.

### DIFF
--- a/src/frontend/swo/sources/socket.ts
+++ b/src/frontend/swo/sources/socket.ts
@@ -5,6 +5,8 @@ import { parseHostPort } from '../../../common';
 import * as vscode from 'vscode';
 import { TextDecoder } from 'util';
 import { setFlagsFromString } from 'v8';
+import * as ChildProcess from 'child_process';
+import * as Path from 'path';
 
 const TimerInterval = 250;
 export class SocketSWOSource extends EventEmitter implements SWORTTSource {
@@ -133,6 +135,39 @@ export class SocketRTTSource extends SocketSWOSource {
             throw e;
         }
     }
+}
+
+export class DefmtSocketRTTSource extends SocketRTTSource {
+    private process: ChildProcess.ChildProcess;
+    private executable: string;
+    private cwd: string;
+
+    protected processData(buffer: Buffer): void {
+        this.process.stdin.write(buffer);
+    }
+
+    public start(timeout = (1000 * 60 * 5)): Promise<void> {
+        this.process = ChildProcess.spawn(
+            'defmt-print',['-e', this.executable], { cwd: this.cwd }
+        );
+
+        this.process.on('error', (e) => {
+            (e as any).message = `Failed to launch defmt-print (is it installed?)`; 
+            this.emit('error', e);
+            this.dispose();
+        });
+
+        this.process.stdout.on('data', (buffer) => this.emit('data', buffer));
+
+        return super.start(timeout);
+    }
+
+    constructor(tcpPort: string, public readonly channel: number, executable: string, wsPath: string) {
+        super(tcpPort, channel);
+
+        this.cwd = wsPath;
+        this.executable = Path.relative(wsPath, executable);
+    }    
 }
 
 export class JLinkSocketRTTSource extends SocketRTTSource {
--- a/package.json
+++ b/package.json
@@ -1134,10 +1134,11 @@
                                                         "type": {
                                                             "enum": [
                                                                 "console",
-                                                                "binary"
+                                                                "binary",
+                                                                "defmt"
                                                             ],
                                                             "default": "console",
-                                                            "description": "'console' with text input/output, 'binary' is for converting byte stream to other data types",
+                                                            "description": "'console' with text input/output, 'binary' is for converting byte stream to other data types, 'defmt' uses defmt-print",
                                                             "type": "string"
                                                         },
                                                         "prompt": {
--- a/src/common.ts
+++ b/src/common.ts
@@ -456,13 +456,14 @@ export class RTTServerHelper {
         });
     }
 
-    public emitConfigures(cfg: RTTConfiguration, obj: EventEmitter): boolean {
+    public emitConfigures(cfg: RTTConfiguration, executable: string, obj: EventEmitter): boolean {
         let ret = false;
         if (cfg.enabled) {
             for (const dec of cfg.decoders) {
                 if (dec.tcpPort || dec.tcpPorts) {
                     obj.emit('event', new RTTConfigureEvent({
                         type: 'socket',
+                        executable: executable,
                         decoder: dec
                     }));
                     ret = true;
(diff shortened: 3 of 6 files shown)


xoviat:
Oh, and defmt-print must be installed for this to work. Maybe add a tooltip for that?

xoviat:
@haneefdm Do you have questions about this?

haneefdm:
Thank you for the PR and I agree it will be very useful for Rust developers Before I look at it, I have a request. I need an example project available on github for me to begin looking at it.

It will be my test case as well. Your github example project should be ready to run with minimal effort on my part (preferably on a Mac). I cannot accept what I cannot test. I am not Rust-educated either.

I you can also update the source code to add a reference to the example github then that would be nice.

Sorry, it took me a long time to get to it. Personal health reasons.

src/frontend/swo/sources/socket.ts
```diff
@@ -135,6 +137,39 @@ export class SocketRTTSource extends SocketSWOSource {
...
+        this.process.stdout.on('data', (buffer) => this.emit('data', buffer));
+
+        return super.start(timeout);
+    }
+
+    constructor(tcpPort: string, public readonly channel: number, executable: string, wsPath: string) {
+        super(tcpPort, channel);
+
+        this.cwd = wsPath;
+        this.executable = Path.relative(wsPath, executable);
```
haneefdm, line 169: Why?

haneefdm:
I lied. I did a cursory review. I don't know what is going on with the `executable` in the `RTTConfigureEvent`.

And, then in the constructor of the `DefmtSocketRTTSource`, you are trying to re-resolve the executable. How can we be assured that the debugger and the RTT decoder is looking at the same executable? Do you need help from the debugger?

Also, I think that the `executable` property in launch.json is optional if you have the `symbolFiles` property.

xoviat:
I'll get back to this in a few days. I need to refresh my memory after taking care of some other stuff.

HaoboGu:
Any updates? I got the same problem, cortex-debug cannot print defmt's RTT log

src/frontend/swo/sources/socket.ts
HaoboGu: @haneefdm I checked the source of defmt and probe-rs. This is because by default, defmt's message is compressed. The detailed info is stored in .elf file, that's why the executable is needed: without .elf, the message cannot be parsed.

See https://github.com/knurling-rs/defmt/blob/main/decoder/src/lib.rs#L157 . The `Table` is what's stored in .elf file

hazer-hazer:
Installed extension from this PR, everything works perfectly, except vscode does not hint to set `defmt` as `type` option in `rttConfig.decoders`.
@Marus Any chance to see this changes in master? 💌

haneefdm:
@xoviat I am still waiting for the PRs original author.

romancardenas:
Any news here? Can I help with any blocker?

wrs:
For those waiting on this, I wrote a comment on the old issue explaining how to get defmt working with the released cortex-debug.

https://github.com/Marus/cortex-debug/issues/755#issuecomment-2561549376

tpilt:
Looking very much forward to this!

> For those waiting on this, I wrote a comment on the old issue explaining how to get defmt working with the released cortex-debug.

This is now further simplified by 'tcp' parameter on defmt-print:
$ defmt-print -w -e target/thumbv7em-none-eabi/debug/blinky tcp

newcomertv:
So the blocker is this? :
> Also, I think that the executable property in launch.json is optional if you have the symbolFiles property.

I think OP is gone. Not a TS dev though if this is the only thing I can maybe have a look and file a PR that can get merged.

GreenCappuccino:
Hello all,

Using the (newer?) `advanced` decoder [described here](https://github.com/Marus/cortex-debug/wiki/SEGGER-RTT-support):

```json
"rttConfig": {
    "enabled": true,
    "address": "auto",
    "decoders": [
        {
            "type": "advanced",
            "decoder": "./rtt-defmt.js",
            "config": {
                "cwd": "${workspaceRoot}",
                "executable": "./target/thumbv8m.main-none-eabihf/debug/interchange",
            },
            "label": "RTT",
            "port": 0,
            "ports": [0]
        },
    ]
},
```

Where I've written an implementation `rtt-defmt.js`:
```js
const ChildProcess = require('child_process');
const Path = require('path');
const vscode = require('vscode');

class DefmtDecoder {
  constructor() {
    this.config = {};
    this.displayOutput = () => {};
    this.graphData = () => {};
    this.process = null;
  }


  init(config, displayOutput, graphData) {
    this.config = config;
    this.displayOutput = displayOutput;
    this.graphData = graphData;

    this.cwd = config.config.cwd || process.cwd();
    if (!config.config.executable) {
      throw new Error(JSON.stringify(config));
    }

    this.executable = Path.resolve(this.cwd, config.config.executable);

    this.process = ChildProcess.spawn(
      'defmt-print',
      ['-e', this.executable],
      { cwd: this.cwd }
    );

    this.process.stdout.on('data', (data) => {
      this.displayOutput(data.toString());
    });

    this.process.stderr.on('data', (data) => {
        this.displayOutput(`stderr: ${data.toString()}`);
    });

    this.process.on('error', (error) => {
        this.displayOutput(`Error: ${error.message}\n`);
    });
  }


  softwareEvent(_port, data) {
    if (!this.process || !this.process.stdin) {
      if (this.outputChannel) {
        this.displayOutput('Error: defmt-print process is not running.\n');
      }
      return;
    }
    let bufferInput = Buffer.isBuffer(data) ? data : Buffer.from(data);
    this.process.stdin.write(bufferInput);
  }

  outputLabel() {
    return this.config.label;
  }

  typeName() {
    return 'defmt-print';
  }
}

module.exports = { default: DefmtDecoder };
```
I have a solution that I prefer, where a new output channel (in this case named `SWO: RTT [type defmt-print]`) is created in the VS code OUTPUT tab with correct `defmt-print` output.

The implementation, when initialized, creates a new `defmt-print` process, which it uses to drive the VS code output channel.

rhuys:
I have the same problem.
probe-rs is not mature enough to be used reliably for debugging, but RTT and DEFMT are essential to work.
Please include DEFMT decoding in cortex-debug.

haneefdm:
Built-in RTT with dfmt-print is coming. This extension is getting a rewrite. Don't want to disturb cortex-debug in a major way as it is used in production by many people. The new extension will be available soon for download here (alpha) and then Beta in the Marketplace. It will be bit of a journey.

https://mcu-debug.github.io/mcu-debug/docs/getting-started
https://mcu-debug.github.io/mcu-debug/docs/tracing/rtt#defmt-support

For now, it is best to use the solution by @GreenCappuccino

rhuys:
@GreenCappuccino thanks for your solution, it works,
but now VSCODE starts a new output tab every time I re-run my application, 
do you have an idea on how to re-use the same tab?

haneefdm:
Removing this #PR since there has been no activity. We have a much better implementation, but it is in a different repo. This extension will not be updated with new features. Only security/bug fixes.

For the next gen extension, See:

https://mcu-debug.github.io/mcu-debug/docs/tracing/rtt#defmt-support

Issue #897, opened 2023-06-08
Adding a debug configuration with .c or .cpp file opened in editor doesn't show cortex-debug option


ashish-gawali-adi:
### Please make you search through our existing [issues](https://github.com/Marus/cortex-debug/issues?q=type:issue) (both open and closed)

**Describe the bug**
Adding a debug configuration with .c or .cpp file opened in editor doesn't show cortex-debug option

**To Reproduce**
1. Open a project with a `.c` or a `.cpp` file
2. Make sure that `.vscode/launch.json` doesn't have any `configurations` defined.
4. Open a `.c` file and select "Run > Add Configurations
![SelectingTheAddConfiguration](https://github.com/Marus/cortex-debug/assets/126178747/17fafc6e-3263-423e-b30e-6080ed59d005)
5. Note the Cortex Debug option is not available in the debugger prompt
![CannotViewCortexDebug Option](https://github.com/Marus/cortex-debug/assets/126178747/5e704ec6-4710-456c-a4cf-5dabb00b427d)
6. Close the `.c` file.
7.  Select "Run > Add Configuration..."
8. The Cortex Debug option is available ![AbleToSeeCortexDebug](https://github.com/Marus/cortex-debug/assets/126178747/b89ac5e3-f523-4b76-aa10-0cd8f5be6380)

**Expected behavior**

[comment]: The expected behavior as I understand it, I should be able to see the `Cortex Debug` option when I have a `.c` or `.cpp` file open, as seen in step 7 and step 8.

**Environment (please complete the following information):**

[comment]: <> Whenever possible, please make sure you are using the latest versions of VSCode and our extension

 - Cortex-Debug Version [1.10.0]
 - OS: [Windows_NT x64 10.0]
 

*Note: We are unlikely to look at the issue if you do not supply this*
```
{
    // Use IntelliSense to learn about possible attributes.
    // Hover to view descriptions of existing attributes.
    // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
    "version": "0.2.0",
    "configurations": [
        {
            "name": "ADI: Cortex Debug with GDB and OpenOCD (ARM Embedded)",
            "cwd": "${config:adi-sdk.debugPath}",
            "executable": "${config:adi-sdk.debugPath}/${config:adi-sdk.programFile}",
            "request": "launch",
            "type": "cortex-debug",
            "runToEntryPoint": "main",
            "showDevDebugOutput": "both",
            "servertype": "openocd",
[...]

PR #898, opened 2023-06-08
Adding languages(c,cpp) for which the debugger option will be prompted.

ashish-gawali-adi:
Resolving issue [#897] by updating the relevant code in pacakge.json. This code adds the `cortex-debug` configuration suggestion for `.c` and `.cpp` files, if there is a need to support more file formats we can add them in the array.

### DIFF
--- a/package.json
+++ b/package.json
@@ -2855,7 +2855,11 @@
                 "program": "./dist/debugadapter.js",
                 "runtime": "node",
                 "type": "cortex-debug",
-                "variables": {}
+                "variables": {},
+                "languages": [
+                    "c",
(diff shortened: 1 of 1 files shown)


haneefdm:
Thanks for the PR. I might also add 'rust' to that list as there are quite a few Rust users.

ashish-gawali-adi:
Hi @haneefdm,
I was wondering when will the latest version of the extension will be available in marketplace. We are using `cortex-debug` as a dependency for our extension.

haneefdm:
I will publish it tomorrow or tonight. What is your extension?

Issue #901, opened 2023-06-22
Getting a default launch configuration


ashish-gawali-adi:
The extension uses `initialConfigurations` to define a default configuration, this option has been deprecated as seen in this [issue](https://github.com/microsoft/vscode/issues/33794). It adds default configuration:
```
{
    "name": "Cortex Debug",
    "cwd": "${workspaceFolder}",
    "executable": "./bin/executable.elf",
    "request": "launch",
    "type": "cortex-debug",
    "runToEntryPoint": "main",
    "servertype": "jlink"
}
```

If you have an extension defining a new configuration of `type: cortex-debug` on creation `launch.json`, you have the default configuration defined in `cortex-debug` as well as the one defined in the extension sitting on top of `cortex-debug`, so you will have `configurations` in `launch.json` somewhat similar to the one below:
```
[
        {
            "name": "Cortex Debug",
            "cwd": "${workspaceFolder}",
            "executable": "./bin/executable.elf",
            "request": "launch",
            "type": "cortex-debug",
            "runToEntryPoint": "main",
            "servertype": "jlink"
        },
        {
            "name": "Debug configuration From new extension",
            "cwd": "${workspaceFolder}",
            "executable": "executable.elf",
            "request": "launch",
            "type": "cortex-debug",
            "runToEntryPoint": "main",
            "servertype": "openocd",
            "showDevDebugOutput": "both",
            "preLaunchCommands": [
                "set logging overwrite on",
                "set logging file debug-arm.log",
                "set logging on",
                "set remotetimeout 60",
                "tbreak abort",
                "tbreak _fatal_error",
                "tbreak _exit"
            ]
        },
]
```

The expected behavior would be to have just one configuration like below:
```
 [
       {
            "name": "Debug configuration From new extension",
            "cwd": "${workspaceFolder}",
            "executable": "executable.elf",
            "request": "launch",
            "type": "cortex-debug",
            "runToEntryPoint": "main",
            "servertype": "openocd",
            "showDevDebugOutput": "both",
            "preLaunchCommands": [
                "set logging overwrite on",
                "set logging file debug-arm.log",
                "set logging on",
                "set remotetimeout 60",
                "tbreak abort",
                "tbreak _fatal_error",
                "tbreak _exit"
            ]
        }
]
```

One possible to solution of this problem to remove `initialConfigurations` from `package.json` and implement function `provideDebugConfigurations` in file `src\frontend\configprovider.ts`.

[...]

PR #902, opened 2023-06-23
Updating method to add initial debug configuration

ashish-gawali-adi:
Updating debug configuration provider, instead of using initialConfigurations in package.json, which is deprecated, using function provideDebugConfigurations. 

The issue targeted is [this](https://github.com/Marus/cortex-debug/issues/901).

### DIFF
--- a/package.json
+++ b/package.json
@@ -2840,17 +2840,6 @@
                         "label": "Cortex Debug: ST-Util"
                     }
                 ],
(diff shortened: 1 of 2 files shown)


ashish-gawali-adi:
@haneefdm please let me know what you think about it.

ashish-gawali-adi:
Hi @haneefdm please let me know what you think of it.

Issue #740, opened 2022-09-08
Missing RTT Timestamp in logfile


J-Rios:
Hello,

Not sure if this is a missing feature or just something that is not working properly on latest versions, but seems that when you setup the RTT configuration in the launch.json file to enable the Timestamp and also log to file features, the timestamp is shown only in the terminal, but not added to the log file.

Here is the launch.json configuration:

```json
{
    "version": "0.2.0",
    "configurations":
    [
        {
            "name": "Debug JLink",
            "type": "cortex-debug",
            "request": "launch",
            "runToEntryPoint": "main",
            "cwd": "${workspaceRoot}",
            "executable": "${workspaceRoot}/build/project_name.elf",
            "svdFile": "${workspaceRoot}/project/svd/STM32L4x1.svd",
            "armToolchainPath": "/usr/bin/",
            "servertype": "jlink",
            "interface": "swd",
            "device": "STM32L431CC",
            "serverpath": "/opt/SEGGER/JLink/JLinkGDBServerCLExe",
            "rttConfig":
            {
                "enabled": true,
                "address": "auto",
                "decoders":
                [
                    {
                        "label": "RTT - STDOUT",
                        "port": 0,
                        "type": "console",
                        "timestamp": true,
                        "logfile": "${workspaceRoot}/build/rtt.log"
                    }
                ]
            }
        }
    ]
}
```

What I get from the VSCode Terminal is:

```text
[2022-09-08T17:18:54.963Z, +251016ms] Hello World
[2022-09-08T17:18:54.984Z, +251037ms] Good bye
```

What is stored in the rtt.log file:

```text
Hello World
Good bye
```

Note: Issue tested in Cortex-Debug v1.6.2 and VSCode 1.71.0 Linux x64.

In case it is a missing feature, could it be considered/discussed as a feature request?
It will be a very nice to have feature.

Best Regards.

haneefdm:
Thanks for reporting this. It was an oversight. We added the timestamp after initial release and forgot all about the timestamp. As the implementation stands right now, this may take some doing. But I will add it to my todo list as I see that as a flaw.

Note that there will be no timestamps for binary type data as raw undecided bytes are written to the file. It will be only for `"type": "console"`

J-Rios:
Hi,

Here you can find a [Pull Request proposal](https://github.com/Marus/cortex-debug/pull/913) for getting this functionality.

Regards!

haneefdm:
[...]

PR #913, opened 2023-07-23
Fix missing timestamp on RTT log file

J-Rios:
Hi,

Here is a possible fix to the missing feature of having timestamps on log file for RTT text based messages.

Notes:

- Related to [this issue](https://github.com/Marus/cortex-debug/issues/740).
- Only applied to text based log, not for binary.
- Functionality developed on specific function to isolate it and avoid affecting any other feature like writing to terminal or binary-write.
- This is my first time touching Typescript, so not sure if it is the best solution in regards time efficiency.

Best Regards :)

### DIFF
--- a/src/frontend/rtt_terminal.ts
+++ b/src/frontend/rtt_terminal.ts
@@ -11,6 +11,7 @@ export class RTTTerminal {
     protected ptyOptions: IPtyTerminalOptions;
     protected binaryFormatter: BinaryFormatter;
     private source: SocketRTTSource;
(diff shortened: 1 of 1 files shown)


haneefdm:
Thank you for creating this PR. I saw your implementation, but I already had an implementation that was sitting in my sandbox. It was not tested, so I didn't release it. In implementing the fix, I realized there is an issue of when the log file is opened and closed and re-opened. Had to address those issues as well.

I am going to close this PR without merging it. Feel free to review/test my implementation

PR #966, opened 2023-12-07
Feature/Line information for customRequest disassemble

itzandroidtab:
This PR adds support for line information about a instruction when running the customRequest `disassemble`. This is so custom plugins can see what line in the source file the instruction belongs to.

This changes `data-disassemble` from mode 2 (disassembly with raw opcodes) to mode 5 (mixed source and disassembly with raw opcodes). This gives line information combined with the raw opcodes (in a slightly different format).

I am currently working on a plugin that adds support for streaming traces (such as the j-trace) and showing the live instruction counters during runtime (still WIP). This would speed up the start time as it would mean I no longer have to call `addr2line` for every instruction.

trace_information

### DIFF
--- a/src/backend/disasm.ts
+++ b/src/backend/disasm.ts
@@ -968,23 +968,41 @@ export class GdbDisassembler {
         const endAddress = symbol.address + symbol.length;
 
         // tslint:disable-next-line:max-line-length
-        const result = await this.miDebugger.sendCommand(`data-disassemble -s ${hexFormat(startAddress)} -e ${hexFormat(endAddress)} -- 2`);
+        const result = await this.miDebugger.sendCommand(`data-disassemble -s ${hexFormat(startAddress)} -e ${hexFormat(endAddress)} -- 5`);
         const rawInstructions = result.result('asm_insns');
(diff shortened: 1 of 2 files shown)


haneefdm:
May I ask why you are using this custom request (which is now obsolete) instead of the official request that is now in the protocol? It already gives you the line info.

https://microsoft.github.io/debug-adapter-protocol/specification#Requests_Disassemble

Also, what extension is this?

itzandroidtab:
Thank you for pointing me in the right direction. The current implementation handles it way better. I did a small test with reading the whole memory region using this method. 

image
(ranges are in next screenshot)

I noticed it only disassembles the vector table and ignores the other addresses. It looks like `findDisasmRanges` only returns one range instead of multiple ranges (in this case clipHigh clips it to the end of the .vectors section). @haneefdm Is this intended?

image

I also tried the previous method I was using where I would request information about a function. It looks like If I send to many requests it crashes GDB. What would be the best time to request this amount of data? Currently it fetches all this data when the plugin gets the 'stopped' event for the first time.
image

This is a side project I am working on [link](https://github.com/itzandroidtab/cortex-trace). Still working on improving the native code that handles the connection with the J-Trace.

Issue #959, opened 2023-11-19
RTT logs are not printed to terminal


HaoboGu:
**Describe the bug**

I'm using openocd + RTT + daplink to debug my board, RTT console doesn't print logs to VSCode terminal. gdb reports it fails to find RTT control block

However, after I paused the program, I do find the control block and the log content which should be printed in the MCU memory:
![image](https://github.com/Marus/cortex-debug/assets/8640918/fb2e2d5a-09b0-4c70-ad54-0b3c30b13107)

**To Reproduce**
1. Start debug session
2. Start the program
3. Check rtt log at Terminal - RTT channel 0

**Expected behavior**

Log should be printed here

**Screenshots**

![image](https://github.com/Marus/cortex-debug/assets/8640918/6ba7cdfb-1a7a-4d38-a9be-c44b66c1deed)

**Environment (please complete the following information):**

 - Cortex-Debug Version (this extension): v1.12.1
 - OS: Windows 11
 - GDB Version: 12.2
 - Compiler Toolchain Version: rustc 1.76.0-nightly 

**Please include `launch.json`**

```json
 {
            "name": "Cortex-Debug",
            "cwd": "${workspaceFolder}",
            "executable": "${workspaceFolder}/target/thumbv7em-none-eabihf/debug/firmware",
            "request": "launch",
            "type": "cortex-debug",
            "servertype": "openocd",
            "showDevDebugOutput": "parsed",
            "runToEntryPoint": "main",
            "configFiles": [
                "${workspaceFolder}/openocd.cfg"
            ],
            "interface": "swd",
            "svdFile": "${workspaceFolder}/STM32H7B0x.svd",
            "rttConfig": {
                "enabled": true,
                "address": "auto",
                "clearSearch": false,
                "polling_interval": 20,
                "rtt_start_retry": 20000,
                "decoders": [
                    {
                        "label": "RTT channel 0",
                        "port": 0,
                        "timestamp": true,
                        "type": "console"
                    }
                ]
            },
        },
```

**Attach text from `Debug Console`**

Please enable debug output in your launch.json (`"showDevDebugOutput": "raw"`). It this is too large, please attach it as a file
```
Cortex-Debug: VSCode debugger extension version 1.12.1 git(652d042). Usage info: https://github.com/Marus/cortex-debug#usage
"configuration": {
    "name": "Cortex-Debug - stm32h7",
    "preLaunchTask": "Flash stm32h7",
[...]

PR #979, opened 2024-02-07
Force tcl connection to IPv4

Slamy:
I first looked for https://github.com/Marus/cortex-debug/issues/959 to find a solution but in the end I compiled this project from source to find the root cause.
After updating on the latest Debian testing, the Rtt connection no longer worked. I wondered why. As it seems, OpenOCD doesn't listen on IPv6 sockets for tcl connection.
Changing to IPv4 allowed to me to continue. This might not be the ideal solution.
I'm creating this PR to at least provide information about this problem.

I don't know why this even started but localhost is resolved to ::1 on my system and not 127.0.0.1 as it was for a long time now. Maybe other applications are suffering with this change I imagine and it will be rolled back...

telnet solved this by first trying ::1 and then retrying with 127.0.0.1 if a connection to localhost shall be established.

### DIFF
--- a/src/openocd.ts
+++ b/src/openocd.ts
@@ -421,7 +421,7 @@ export class OpenOCDServerController extends EventEmitter implements GDBServerCo
                 const tclPortName = createPortName(0, 'tclPort');
                 const tclPortNum = this.ports[tclPortName];
                 const obj = {
-                    host: 'localhost',
+                    host: '127.0.0.1',
                     port: tclPortNum
                 };
                 this.tclSocket = net.createConnection(obj, () => {


haneefdm:
I have to review the impact of this across all OSes. We have hundreds of thousands of users and there could be a fierce blowback. But I am busy for the next month.

Not many support IPv6. Question is, why is your Debian configured for IPv6 for your local network? Was this your choice? Or the default?

Windows for example will support both IPv6 and IPv4 and localhost still resolves to IPv4.

Also, why not address this with OpenOCD?

Slamy:
Alright. I've found the culprit.
https://sourceforge.net/p/openocd/tickets/424/

Please leave this PR open for now as some sort of guidance for fellow developers who might stumble onto this problem.
If OpenOCD solves this, I'll provide you with feedback.

haneefdm:
A similar PR has already been merged, Yes, this is openocd-specific

PR #990, opened 2024-02-27
Adds possibility to change internal gdb server port

KGU-SRE:
this is useful when local machine is configured to use 50000- 52000 port range for other things

### DIFF
--- a/package.json
+++ b/package.json
@@ -560,6 +560,11 @@
                                 "items": "string",
                                 "description": "Additional GDB Commands to be executed at the beginning of the restart sequence (after interrupting execution)."
                             },
+                            "gdbPort": {
+                                "default": 50000,
+                                "type": "number",
+                                "description": "Port to use for internal GDB server"
+                            },
                             "postRestartCommands": {
                                 "default": [],
                                 "type": "array",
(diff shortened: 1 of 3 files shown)


KGU-SRE:
Moved PR to separate branch

KGU-SRE:
Hi, this fixed port issue is currently breaking using the internal "pe" GDB server on our machines. The current workaround is to either launch the PEMicro GDB server externally (default port 7224) or to downgrade to cortex-debug v1.4.4. Both of which come with some drawbacks

@haneefdm @Marus how could we fast-track this fix so all of the team can simply use the plugin without modifications?

cpath-star:
Just wanted to say this would be a welcome change. Currently running into port conflicts with no easy way to resolve them

haneefdm:
First of all, there should be no port conflicts ... as we are searching for open ports. All 2000 ports were taken? We want to fix the root cause. I don't see an issue submitted for this.

haneefdm:
Even if there is a port conflict, the would be the wrong fix The ports are not just for GDB but also for gdb-servers and internal use. I am rejecting this PR.

KGU-SRE:
@haneefdm how would you propose to fix the issue with a system where the port range for GDB servers is blocked/used for other purposes?

PR #1003, opened 2024-04-04
Detailed explanation in debug_arguments.md

ECMGS:
Hi,

I was trying to use cortex-debug with a custom version of qemu and I wanted to set a custom executable path for qemu, however I was having issues doing so. I think that adding this note could be helpful for other people in the future.

Best regards.


PhilippHaefele:
Hi @ECMGS,

Thanks for contributing 🙂.

As the debug_attributes.md is generated from package.json you should, change it there and regenerate with the python script. Be aware that (most) configuration options are doubled up (for launch and attach).

Not sure if we should really add additional text to the already very long one.

But let @haneefdm decide

ECMGS:
Hi @PhilippHaefele,

I've updated package.json as you commented and then updated debug_attributes.ms by using the script in dist/docgen.js- I've also tried to condensed the explanation a bit.

Hope this helps!

Issue #999, opened 2024-03-22
No RTT connect retry with OpenOCD


kossnikita:
**Describe the bug**

cortex-debug try to start RTT before RTT initialization, so OpenOCD unable to find RTT block. But there's no connect retry after that. Even with `rtt_start_retry > 0`

But I able to start RTT with telnet `rtt start` after `SEGGER_RTT_ConfigUpBuffer()` during debug session and this works.

Also I have similar project with same setup (board + stlink) and this works perfectly on same host with same `launch.json`. Can't say what difference between projects.

**Expected behavior**

cortex-debug should retry connect to RTT if `rtt_start_retry > 0`

**Environment:**

 - Cortex-Debug 1.12.1 git(652d042)
 - OS: Windows_NT x64 10.0.22621
 - GDB Version: Open On-Chip Debugger 0.11.0-00006-gbec841d3f-dirty
 - Compiler Toolchain Version: Arm GNU Toolchain 13.2.rel1 (Build arm-13.7)) 13.2.90.20231008-git
 - 
**launch.json**

```json
{
    "version": "0.2.0",
    "configurations": [
        {
            "type": "cortex-debug",
            "name": "Cortex Debug",
            "cwd": "${workspaceFolder}",
            "serverpath": "D:/ARM/tool-openocd-at32-0.1100.220911/bin/openocd.exe",
            "executable": "${workspaceFolder}/build/firmware",
            "request": "launch",
            "servertype": "openocd",
            "svdFile": "${workspaceFolder}/peripherals.svd",
            "showDevDebugOutput": "raw",
            "runToEntryPoint": "main",
            "configFiles": [
                "interface/stlink.cfg",
                "target/at32f421xx.cfg"
            ],
            "rttConfig": {
                "enabled": true,
                "address": "auto",
                "clearSearch": false,
                "rtt_start_retry": 1000,
                "decoders": [
                    {
                        "port": 0,
                        "type": "console",
                    },
                ]
            },
        }
    ]
}
```

**Attach text from `Debug Console`**

```
Cortex-Debug: VSCode debugger extension version 1.12.1 git(652d042). Usage info: https://github.com/Marus/cortex-debug#usage
"configuration": {
    "type": "cortex-debug",
    "name": "Cortex Debug",
    "cwd": "D:\\ARM\\oled-test",
    "serverpath": "D:/ARM/tool-openocd-at32-0.1100.220911/bin/openocd.exe",
[...]

PR #1018, opened 2024-05-21
Fix the openocd tcl socket connection problem

hulryung:
The openocd tcl socket connection problem is caused by the host name 'localhost' which is resolved to '::1' on some systems. This patch changes the host name to '127.0.0.1' to avoid the problem.

This patch is related these issues:
https://github.com/Marus/cortex-debug/issues/999


haneefdm:
I dont see how this is related to #999. Can you please submit a new issue so I can understand the platform and environment where 'localhost' is resolving to '::1' and what does not work.

haneefdm:
Yes, I see now that this is openocd specific and I can approve this PR.

Issue #1026, opened 2024-06-16
Project status, is it still actively developed/maintained?


daantimmer:
I created an issue about 4mo ago. And noticed there was a severe lack of activity. 

Marus, the original author hasn't provided any work in a long time. And the only work that was done was by another individual who seems to have disappeared as well. 

The last commit was 4mo ago, and responses to (new) issues also seem to have stopped around that same time. 

It is slowly making me a bit nervous, as this is a highly dependent upon extension for any embedded development for vscode.

Marus:
I wish I had better news for you in terms of more active development - unfortunately I haven't had time to be involved myself in recent years. I don't actually work in the embedded space - so Cortex-Debug was primarily developed to support my hobbyist development - unfortunately I haven't been doing much in that area recently - which has limited the time I have to support and develop this project.

@haneefdm had been doing most of the maintenance in the last few years - but I haven't heard from him recently - so possible that they don't have time to deal with it on top of their other commitments. 

Am always happy to consider getting other contributors that would like to take on a more major role setup.

I will try to make some time to at least look at open PRs to see if I think they can be safely merged into the project and released.

haneefdm:
Please see #1027. I also updated a few PRs and still waiting for replies from the authors.

ssimek:
@Marus @haneefdm if you guys are still interested, I'm open to discussion about picking up the maintenance of this extension

PR #1052, opened 2024-10-12
Proper SWO support for the Black Magic Probe

ssimek:
Contains the following changes

* slight modification of the deployment model so the binary dependencies are included directly in the package, the same way they are in e.g. arm.device-manager extension
* something must have changed in the meantime with Electron, because the original prebuilt binaries seem to work on all platforms without recompilation
  * tested on macOS, Win and Linux (ARM64) - everything works except SerialPort on Windows ARM64, but that wouldn't work with the old approach just the same
* a new `UsbSWOSource` has been added that reads a raw bulk endpoint
  * it autodetects the probe based on productName and interfaceName, both can be overridden

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,7 +1,7 @@
 out
 /node_modules
 /webview/node_modules
-/binary_modules/*/node_modules
+/binary_modules/**/node_modules
 *.vsix
 .vscode-test
 .DS_Store
--- a/.vscodeignore
+++ b/.vscodeignore
@@ -10,6 +10,6 @@ tsconfig.json
 vsc-extension-quickstart.md
 webpack.config.js
 node_modules
-binary_modules/electron*
+binary_modules
 
 !node_modules/@vscode/webview-ui-toolkit
--- a/binary_modules/package-lock.json
+++ b/binary_modules/package-lock.json
@@ -9,7 +9,8 @@
       "version": "1.0.0",
       "license": "MIT",
       "dependencies": {
-        "serialport": "^10.4.0"
+        "serialport": "^10.4.0",
+        "usb": "^2.14.0"
       },
       "devDependencies": {
         "electron-rebuild": "^3.2.8"
@@ -334,6 +335,12 @@
       "dependencies": {
         "@types/node": "*"
       }
+    },
+    "node_modules/@types/w3c-web-usb": {
+      "version": "1.0.10",
+      "resolved": "https://registry.npmjs.org/@types/w3c-web-usb/-/w3c-web-usb-1.0.10.tgz",
+      "integrity": "sha512-CHgUI5kTc/QLMP8hODUHhge0D4vx+9UiAwIGiT0sTy/B2XpdX1U5rJt6JSISgr6ikRT7vxV9EVAFeYZqUnl1gQ==",
+      "license": "MIT"
     },
     "node_modules/abbrev": {
       "version": "1.1.1",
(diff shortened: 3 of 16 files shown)


haneefdm:
Question: Did you just package the serial port binary extension with this extension?

If so, I cannot accept that. We used to do that but last year we removed it because of numerous compatibility issues with Node and VSCode updates. We had to keep updating our extension constantly when VSCode updated and we also had to keep multiple versions of the binary serial port package.

This is a lot of constant work to keep releasing new versions even when nothing changed in our code or in the serial port code.

Instead, we relied on Microsoft's serial port facility. We should probably make that a requirement. I don't have the resources Microsoft/ARM does.

https://marketplace.visualstudio.com/items?itemName=ms-vscode.vscode-serial-monitor

ssimek:
I did, and I have also extensively tested it and it seems to work reliably. I took inspiration from other extensions that do exactly the same (e.g. ARM Device Manager, or the Microsoft one you used to "borrow" libusb from).

haneefdm:
Okay, but not sure you know what happens with every release of VSCode. Testing with one version does not suffice. We have to rebuild and re-release for all platforms every time VSCode updates Electron. Happens several times a year.

ssimek:
True and I don't insist on putting this in. I'm using my version since October with no ill effects even after three VS minor updates. I'm quite happy waiting for a few more months.
Also really glad to see you guys back! Funny thing actually - I started working on publishing my fork *literally* an hour ago or so.

haneefdm:
The trouble happens when VSCode switches verion of Electron. All  binary extensions break and that the nature of Electron. Not every update of VSCode updates their Electron/Node dependency. That is why it is random and unpredictable. I hate to break other peoples workflow because of my extension.

We should focus on supporting SWO first. Bundling the serial port should be a separate (and independent) topic from SWO support.

ssimek:
I respectfully disagree. While you technically can keep the hack in for serial port, it would be untenable for libusb (I spent an unhealthy amount of time trying to make it work only to find out it really seems to be possible to include unmodified native extensions). Also, I'm not an expert in native node dependencies, but I believe something has to have changed in the meantime - both libusb and serialport use NAPI which claims to be stable and version independent.

haneefdm:
The N-API is good news. I will look into it. I was also getting warnings from the marketplace that my extension is too large and that I should split it up. This was when we were releasing 2-3 versions of binary serial port packages for three OSes.

How big is the packaged extension with serial-port bundled?

What is your objection to splitting up the SWO support from binary package bundling?

I would like to clone it and see what you did. Did you use/modify the existing script to build the packages?

haneefdm:
Thanks for refactoring the gdb init scripts

haneefdm:
I am looking at your changes. Oh, I didn't understand what you did. Earlier, we had to compile and bundle the serial port ourselves manually. I used to spend 2-3 days doing this several times a year. Now, you are using it as a normal dependency via package.json?

ssimek:
All I did is in this PR :) See the change in the [vscode:prepublish npm task](https://github.com/Marus/cortex-debug/pull/1052/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519), it just runs `npm ci` in the `binary_modules` and then copies the installed node_modules to dist

haneefdm:
Okay, ELI5 to me. If all the dist has is just javascript file, why do we need to build and put it in the dist directory?

ssimek:
Not just javascript, it gets the entire node_modules for the dependencies containing native components. They also have to be excluded from webpacking (`externals` in webpack.config.js).

The entire output tree looks like this (the sizes are misreported as bytes instead of KB for some reason).

```
[4.5K]  .
└── [4.4K]  extension
    ├── [2.5K]  dist
    │   └── [2.2K]  node_modules
    │       ├── [ 512]  @serialport
    │       ├── [  96]  @types
    │       ├── [ 192]  debug
    │       ├── [ 192]  ms
    │       ├── [ 288]  node-gyp-build
    │       ├── [ 192]  serialport
    │       └── [ 480]  usb
    ├── [ 416]  images
    ├── [ 448]  node_modules
    │   └── [ 352]  @vscode
    │       └── [ 256]  webview-ui-toolkit
    ├── [ 160]  resources
    ├── [ 192]  support
    └── [ 128]  syntaxes
```

As you can see, what's new is the `dist/node_modules` folder, containing untouched versions downloaded via npm. This is very similar to what the original patch was doing - pointing to a very similar node_modules folder inside another extension.

haneefdm:
> Also really glad to see you guys back! Funny thing actually - I started working on publishing my fork literally an hour ago or so.

Coincidental. Btw, there are no 'guys'. Just me. I am monitoring PRs and Issues and addressing them when I can. There are several changes sitting on my laptop that I haven't merged yet. Some of it had to do with changes in VSCode that we can take advantage of.

ssimek:
> Coincidental. Btw, there are no 'guys'. Just me. I am monitoring PRs and Issues and addressing them when I can. There are several changes sitting on my laptop that I haven't merged yet. Some of it had to do with changes in VSCode that we can take advantage of.

Oh. I thought Marcel is still around, based on #1026 (I tried to reach out to you both there some time ago to see if I can help). That said, is there any way I can help?

ssimek:
Sounds great, feels like I can actually fill a complementary space - I'm mostly on macOS, using all kinds of probes except OpenOCD (had terrible experience with it about 10-15 years ago performance-wise, leading to me writing my own tool for driving SWD/SWO over FTDI, not sure what the situation is now, maybe I'll look into again).

I've been using and loved cortex-debug since its beginnings, but I've been away from microcontroller development for several years now. I'm planning to get back to it and invest quite a bit of time this year. A lot of what I'd like to work on is motivated by "selfish" reasons (adding missing functionality I personally need/want), but I can certainly help with existing issues and communicate with other users.

daantimmer:
I've got nothing of value to contribute besides a big thank you for spending your own time on this project.

PR #1072, opened 2025-01-13
SWO support when using ST-LINK GDB server

ssimek:
Adds the ability to use the raw SWO stream from the ST-LINK debugger.

One important change is waiting for SWO connection during startup - this seems to be necessary because starting the actual launch completely confuses ST-LINK's GDB and leaves it in a semi-frozen state. I'm sure this change should be beneficial for other adapters as well as I've seen SWO randomly not working from time to time and this is a possible explanation.

For the future, we could consider refactoring the GDB server controllers for better code sharing, there is not much about SWO init that is server specific, other than an occasional `monitor` command.


haneefdm:
Can you please update the CHANGELOG.md as well? Just something brief for anything that might be user-visible.

haneefdm:
Especially for new features.

PR #1076, opened 2025-01-16
chore: update all NPM packages to latest

ssimek:
A surprisingly small change needed to get all our dependencies to their latest versions. Everything seems to be working fine.

The only dependency I've left untouched is vscode since I have no idea what kind of old version support policy do we have.


haneefdm:
It appears that the Build failed during CI run. See above.

FYI: We are moving to version  1.13.x of cortex debug. Meaning it may have breaking changes or a LOT of changes. About VSCode, we try to stay update to the latest when we have some free time. Once or twice a year. About our verioning methodology.

https://github.com/Marus/cortex-debug?tab=readme-ov-file#release-versioning

We can push 1.13.x to the marketplace and only people who subscribe will be affected.

PR #1081, opened 2025-01-19
fix: incorrect RingBuffer import after no-require-imports

ssimek:
@haneefdm sorry, I have accidentally injected an issue into SWO startup when fixing some of the simpler rule violations.


haneefdm:
glad you caught it

PR #1095, opened 2025-03-03
Fix path normalization on windows.

PILIX123:
The `path.normalize()` function returns a path with the `\\` on windows. so i did a check after the normalize to replace the `\\` if the platform is windows.

### DIFF
--- a/src/gdb.ts
+++ b/src/gdb.ts
@@ -476,6 +476,9 @@ export class GDBDebugSession extends LoggingDebugSession {
 
         if (args.executable && !path.isAbsolute(args.executable)) {
             args.executable = path.normalize(path.join(args.cwd, args.executable));
+            if(os.platform().startsWith("win")){
+                args.executable = args.executable.replace(/\\/ig,"/");
+            }
         }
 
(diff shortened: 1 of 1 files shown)


PILIX123:
#1101

src/gdb.ts
```diff
@@ -476,6 +476,9 @@ export class GDBDebugSession extends LoggingDebugSession {
 
         if (args.executable && !path.isAbsolute(args.executable)) {
             args.executable = path.normalize(path.join(args.cwd, args.executable));
+            if(os.platform().startsWith("win")){
```
haneefdm, line 479: Why the `.startsWith`. Why not compare it to 'win32'. That is officially what it is. Also make sure you run 'npm run lint'

haneefdm:
See https://nodejs.org/api/os.html#osplatform

PILIX123:
@haneefdm Will do tonight. Thanks for the pointers i didnt know that win32 was official and win64 doesnt exist.

PILIX123:
@haneefdm I fixed the issues, sorry for the delay

PR #1097, opened 2025-03-08
always report gdb launch errors during launch (fixes #1096)

### LINKED ISSUE #1096: Debug session silently fails to start if executable exists but is not the correct format
**Describe the bug**

No error dialog or notification is displayed if `-file-exec-and-symbols` command fails, for example due to the binary being in the wrong format.

Strangely, if certain delays are artificially induced in `cortex-debug` with the debugger, it's possible to get the expected dialog box to appear in the client project.

This may not seem like a big deal, but it's hindering the usability of a project I'm setting up which can build targets for different platforms using different CMake toolchain presets and uses `"executable": "${command:cmake.launchTargetPath}"` to ensure the executable was built from the latest sources before launching the debug session.

**To Reproduce**

1. In `launch.json`, set `"executable"` to a path that exists, but isn't a valid ARM binary
2. Start debug session
3. **The debug session tries to start, but fails immediately and silently**
4. Check the *DEBUG CONSOLE*
5. **See no clear indication of what went wrong**
6. Try again with `"showDevDebugOutput": "raw"`
7. **see the below lines in the *DEBUG CONSOLE*:**
```
10-file-exec-and-symbols "/path/to/binary"
-> 10^error,msg="\"/path/to/binary\": not in executable format: file format not recognized"
```
8. Run `cortex-debug` server using the **Debug Server** launch configuration
9. Put a breakpoint on the first line of `src/backend/server.ts:GDBServer.onStdout()`
10. Start the `cortex-debug` server debug session
11. Add `"debugServer":4711` to the firmware project's `launch.json` configuration.
12. Start an embedded firmware project debug session
13. Whenever the breakpoint is hit in `cortex-debug`, **Continue (F5)** quickly afterward (if you wait too long, the client debug session will time out, which isn't the error we're trying to see)
14. For whatever reason, with these artificial delays added, **the client project then pops up a "Failed to generate gdb commands" error dialog** instead of failing silently

**Expected behavior**

An error dialog saying *Failed to generate gdb commands: "/path/to/binary": not in executable format: file format not recognized (from file-exec-and-symbols "/path/to/binary")* followed by a stack trace.

**Screenshots**

The error dialog, which only pops up if the debugger is used to create artificial delays:

Image

**Environment:**

 - Cortex-Debug Version: v1.12.1
 - OS: Linux Manjaro 25.0.0 (rolling release)
 - GDB Version: 15.2
 - Compiler Toolchain Version: LLVM-ET-Arm clang version 18.1.3, Target: arm-none-unknown-eabi

**launch.json**

```json
{
    "version": "0.2.0",
    "configurations": [
        {
            "name": "Debug Native Tests",
            "type": "lldb",
            "request": "launch",
            "program": "${command:cmake.launchTargetPath}",
            "args": [],
            "cwd": "${workspaceFolder}",
            "preRunCommands": [
                "breakpoint name configure --disable cpp_exception"
            ],
        },
        {
            "name": "Debug with J-Link (JTAG)",
            "type": "cortex-debug",
            "request": "launch",
            "cwd": "${workspaceFolder}",
            "runToEntryPoint": "main",
            "showDevDebugOutput": "raw",
            "servertype": "jlink",
            "interface": "jtag",
            "device": "rm44l520",
            "serverArgs": ["-speed", "4000"],
            "liveWatch": {
                "enabled": true,
                "samplesPerSecond": 4
            },
            "svdFile": "rm44l520.svd",
       },
    ]
}
```

**Text from `Debug Console`**

```
Cortex-Debug: VSCode debugger extension version 1.12.1 git(652d042+dirty). Usage info: https://github.com/Marus/cortex-debug#usage
"configuration": {
    "name": "Debug with J-Link (JTAG)",
    "cwd": "/home/malsyned/work/preferred/cortex-r-template",
    "executable": "/home/malsyned/work/preferred/cortex-r-template/out/build/debug/cortex-r-template-CppUTest",
    "request": "launch",
    "type": "cortex-debug",
    "runToEntryPoint": "main",
    "showDevDebugOutput": "raw",
    "servertype": "jlink",
    "interface": "jtag",
    "device": "rm44l520",
    "serverArgs": [
        "-speed",
        "4000"
    ],
    "liveWatch": {
        "enabled": true,
        "samplesPerSecond": 4
    },
    "svdFile": "/home/malsyned/work/preferred/cortex-r-template/rm44l520.svd",
    "debugServer": 4711,
    "__configurationTarget": 6,
    "gdbServerConsolePort": 55878,
    "pvtAvoidPorts": [],
    "chainedConfigurations": {
        "enabled": false
    },
    "debuggerArgs": [],
    "swoConfig": {
        "enabled": false,
        "decoders": [],
        "cpuFrequency": 0,
        "swoFrequency": 0,
        "source": "probe"
    },
    "rttConfig": {
        "enabled": false,
        "decoders": []
    },
    "graphConfig": [],
    "preLaunchCommands": [],
    "postLaunchCommands": [],
    "preAttachCommands": [],
    "postAttachCommands": [],
    "preRestartCommands": [],
    "postRestartCommands": [],
    "preResetCommands": [],
    "postResetCommands": [],
    "toolchainPrefix": "arm-none-eabi",
    "extensionPath": "/home/malsyned/.vscode/extensions/marus25.cortex-debug-1.12.1",
    "registerUseNaturalFormat": true,
    "variableUseNaturalFormat": true,
    "pvtVersion": "1.12.1",
    "__sessionId": "68152115-56a2-453a-bfa7-e63c3a3fe8ef",
    "pvtShowDevDebugOutput": "raw"
}
Reading symbols from arm-none-eabi-objdump --syms -C -h -w /home/malsyned/work/preferred/cortex-r-template/out/build/debug/cortex-r-template-CppUTest
Reading symbols from arm-none-eabi-nm --defined-only -S -l -C -p /home/malsyned/work/preferred/cortex-r-template/out/build/debug/cortex-r-template-CppUTest
Launching GDB: arm-none-eabi-gdb -q --interpreter=mi2
1-gdb-version
Launching gdb-server: JLinkGDBServer -singlerun -nogui -if jtag -port 50000 -swoport 50001 -telnetport 50002 -device rm44l520 -speed 4000
    Please check TERMINAL tab (gdb-server) for output from JLinkGDBServer
Finished reading symbols from objdump: Time: 30 ms
Finished reading symbols from nm: Time: 28 ms
'objdump' exited with a nonzero exit status 1, null
'nm' exited with a nonzero exit status 1, null
-> =thread-group-added,id="i1"
-> ~"GNU gdb (GDB) 15.2\n"
-> ~"Copyright (C) 2024 Free Software Foundation, Inc.\n"
-> ~"License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>\nThis is free software: you are free to change and redistribute it.\nThere is NO WARRANTY, to the extent permitted by law."
-> ~"\nType \"show copying\" and \"show warranty\" for details.\n"
-> ~"This GDB was configured as \"--host=x86_64-pc-linux-gnu --target=arm-none-eabi\".\n"
-> ~"Type \"show configuration\" for configuration details.\n"
-> ~"For bug reporting instructions, please see:\n"
-> ~"<https://www.gnu.org/software/gdb/bugs/>.\n"
-> ~"Find the GDB manual and other documentation resources online at:\n    <"
-> ~"http://www.gnu.org/software/gdb/documentation/>.\n\n"
-> ~"For help, type \"help\".\n"
-> ~"Type \"apropos word\" to search for commands related to \"word\".\n"
-> 1^done
2-gdb-set mi-async on
-> 2^done
3-interpreter-exec console "set print demangle on"
-> 3^done
4-interpreter-exec console "set print asm-demangle on"
-> =cmd-param-changed,param="print asm-demangle",value="on"
-> 4^done
5-enable-pretty-printing
-> 5^done
6-interpreter-exec console "source /home/malsyned/.vscode/extensions/marus25.cortex-debug-1.12.1/support/gdbsupport.init"
-> 6^done
7-interpreter-exec console "source /home/malsyned/.vscode/extensions/marus25.cortex-debug-1.12.1/support/gdb-swo.init"
-> =cmd-param-changed,param="language",value="c"
-> =cmd-param-changed,param="language",value="auto"
-> 7^done
8-interpreter-exec console "set output-radix 0xa"
-> ~"Output radix now set to decimal 10, hex a, octal 12.\n"
Output radix now set to decimal 10, hex a, octal 12.
-> 8^done
9-interpreter-exec console "set input-radix 0xa"
-> ~"Input radix now set to decimal 10, hex a, octal 12.\n"
Input radix now set to decimal 10, hex a, octal 12.
-> 9^done
10-file-exec-and-symbols "/home/malsyned/work/preferred/cortex-r-template/out/build/debug/cortex-r-template-CppUTest"
-> 10^error,msg="\"/home/malsyned/work/preferred/cortex-r-template/out/build/debug/cortex-r-template-CppUTest\": not in executable format: file format not recognized"
```

**Additional context**

Without slowing down the debug server artificially with breakpoints, the client debug session exits silently, but the server debug console provides a little more information:

```
/usr/bin/node --nolazy --trace-warnings ./dist/debugadapter.js --server=4711
waiting for debug protocol on port 4711
>> accepted connection from client
Process exited with code 1
Uncaught MIError MIError
    at errReport (/home/malsyned/packages/cortex-debug/src/backend/mi2/mi2.ts:1049:32)
    at <anonymous> (/home/malsyned/packages/cortex-debug/src/backend/mi2/mi2.ts:1073:21)
    at <anonymous> (/home/malsyned/packages/cortex-debug/src/backend/mi2/mi2.ts:284:52)
    at onOutput (/home/malsyned/packages/cortex-debug/src/backend/mi2/mi2.ts:259:15)
    at stdout (/home/malsyned/packages/cortex-debug/src/backend/mi2/mi2.ts:214:18)
    at emit (<node_internals>/events:507:28)
    at addChunk (<node_internals>/internal/streams/readable:559:12)
    at readableAddChunkPushByteMode (<node_internals>/internal/streams/readable:510:3)
    at <anonymous> (<node_internals>/internal/streams/readable:390:5)
    at onStreamRead (<node_internals>/internal/stream_base_commons:189:23)
    at callbackTrampoline (<node_internals>/internal/async_hooks:130:17)
```

I suspect the issue is with the way GDB and the JLinkGDBServer are being launched in parallel using promises, although I can't pinpoint exactly why that's a problem.

One way to fix this issue is to modify `GDBDebugSession.processLaunchAttachRequest()` by moving the line
```typescript
            const gdbPromise = this.startGdb(response);
```
inside the block that starts:
```typescript
                this.server.init().then(async (started) => {
```

I didn't submit a PR for that, though, because it looks to me like these are done asynchronously in the interest of startup time, and I don't know how this change will affect that (although it seems to start up pretty quickly for me either way).

malsyned:
I'm not well-enough versed in JavaScript async & Promises and the way they interact to be sure about this, but as far as I can tell, the problem seems to be that if the Node interpreter runs out of events to process and still has some rejected promises with no reject handlers registered, it considers that an error and crashes with a stack trace. This patch makes sure that control is never handed back to the event loop with no reject handler registered for `gdbPromise`.

I'm not 100% confident in my diagnosis or solution, but it works for me, both with and without the breakpoint I described in #1096. If someone with more understanding of the interaction between async/await and Promises wants to steer me in a different direction, I'm interested.

### DIFF
--- a/src/gdb.ts
+++ b/src/gdb.ts
@@ -575,7 +575,9 @@ export class GDBDebugSession extends LoggingDebugSession {
                 return doResolve();
             }
             const symbolsPromise = this.loadSymbols();      // This is totally async and in most cases, done while gdb is starting
+            let gdbPromiseAsyncErr;
             const gdbPromise = this.startGdb(response);
+            gdbPromise.catch((err) => { gdbPromiseAsyncErr = err });
             this.usingParentServer = this.args.pvtMyConfigFromParent && !this.args.pvtMyConfigFromParent.detached;
             this.getTCPPorts(this.usingParentServer).then(async () => {
                 await this.serverController.allocateRTTPorts();     // Must be done before serverArguments()
@@ -648,6 +650,8 @@ export class GDBDebugSession extends LoggingDebugSession {
                         // 3. Found free TCP ports and launched gdb-server
                         // 4. Finished reading symbols from objdump and nm
                         const showTimes = this.args.showDevDebugOutput && this.args.showDevDebugTimestamps;
+                        if (gdbPromiseAsyncErr)
+                            throw gdbPromiseAsyncErr;
                         await gdbPromise;
                         if (showTimes) { this.handleMsg('log', 'Debug Time: GDB Ready...\n'); }
 
--- a/src/gdb.ts
+++ b/src/gdb.ts
@@ -577,7 +577,9 @@ export class GDBDebugSession extends LoggingDebugSession {
             const symbolsPromise = this.loadSymbols();      // This is totally async and in most cases, done while gdb is starting
             let gdbPromiseAsyncErr;
             const gdbPromise = this.startGdb(response);
-            gdbPromise.catch((err) => { gdbPromiseAsyncErr = err });
+            gdbPromise.catch((err) => {
+                gdbPromiseAsyncErr = err;
+            });
             this.usingParentServer = this.args.pvtMyConfigFromParent && !this.args.pvtMyConfigFromParent.detached;
             this.getTCPPorts(this.usingParentServer).then(async () => {
                 await this.serverController.allocateRTTPorts();     // Must be done before serverArguments()


haneefdm:
I am in the middle of something that will complicate this merge. This is a good solution and I will merge the changes when things are stable.

However, lint is not passing. You can do a `npm run lint`.

haneefdm:
There may be another problem. I see a stack trace in #1096. I think we should look at why that is happening. We may be masking that bug by doing what you did in this PR. We should merge this change but also look at why we are crashing.

If you can help, that would be great.

malsyned:
> There may be another problem. I see a stack trace in #1096. I think we should look at why that is happening. We may be masking that bug by doing what you did in this PR. We should merge this change but also look at why we are crashing.
> 
> If you can help, that would be great.

Happy to help. I don't think the stack trace is evidence of anything else broken, but I agree that it would be nice to get rid of it.

It's happening because:
1. GDB replies to `-file-exec-and-symbols` with an `error` response.
2. The `errReport()` closure in `sendOneCommand()` calls `reject()` with a new `MIError`
3. The `try/catch` block that starts a little before `await gdbPromise` catches this `MIError` which is thrown either from the `await` line or the new `throw` right before it that this PR adds.

I think this is all working well enough. The problem is that the `catch` block in question is trying to catch all errors from a big block of code, and is written fairly generically to attach a stack trace to whatever is thrown. Does it make sense to check specifically for `MIError` and send a `launchErrorResponse` that doesn't include the stack trace?

Something like:
```diff
@@ -688,9 +690,14 @@ export class GDBDebugSession extends LoggingDebugSession {
                             commands.push(...this.args.postLaunchCommands.map(COMMAND_MAP));
                         }
                     } catch (err) {
-                        const msg = err.toString() + '\n' + err.stack.toString();
-                        this.sendEvent(new TelemetryEvent('Error', 'Launching GDB', `Failed to generate gdb commands: ${msg}`));
-                        this.launchErrorResponse(response, 104, `Failed to generate gdb commands: ${msg}`);
+                        let msg: string;
+                        if (err instanceof MIError) {
+                            msg = `Failed to initialize GDB: ${err.message}\n\ncommand: ${err.source}`;
+                        } else {
+                            msg = 'Failed to generate gdb commands: ' + err.toString() + '\n' + err.stack.toString();
+                        }
+                        this.sendEvent(new TelemetryEvent('Error', 'Launching GDB', msg));
+                        this.launchErrorResponse(response, 104, msg);
                         return doResolve();
                     }

```

Giving an error popup like this:
image

I don't have strong opinions about the exact wording of the error. I included the command that resulted in the error, but I'm not sure that actually adds anything of value. I'm also not completely clear on the significance of the `104` error code, or whether it's appropriate to use it for both kinds of messages there. Can you give any guidance on that?

PR #1108, opened 2025-03-31
Enable/Disable Hex mode: modify User preference, not Workspace

### LINKED ISSUE #1107: Enable Hex mode button: Modify `cortex-debug.variableUseNaturalFormat` in User settings, not Workspace settings
**Is your feature request related to a problem? Please describe.**
On our project, the workspace's `.vscode/settings.json` is tracked in Git, and so every time a developer toggles "Enable Hex mode", it creates noise and/or conflicts in Git.

**Describe the solution you'd like**
Make the "Enable Hex mode" button in the WATCH and CORTEX LIVE WATCH panes modify the user's `settings.json` instead of the workspace's `.vscode/settings.json`.

**Describe alternatives you've considered**
We have considered not tracking `.vscode/settings.json` in Git, instead having each user create it from a template or list of instructions, but there's just too much project policy contained in it, and additions are made too frequently, for that to be convenient. We've also considered training all of the developers to add hunks of `settings.json` interactively and individually to their commits, but I consider this too fiddly and error-prone to require of everyone on the team.

malsyned:
Fixes #1107

### DIFF
--- a/src/frontend/extension.ts
+++ b/src/frontend/extension.ts
@@ -339,7 +339,7 @@ export class CortexDebugExtension {
 
         vscode.commands.executeCommand('setContext', `cortex-debug:${CortexDebugKeys.VARIABLE_DISPLAY_MODE}`, newVal);
         try {
-            config.update(CortexDebugKeys.VARIABLE_DISPLAY_MODE, newVal);
+            config.update(CortexDebugKeys.VARIABLE_DISPLAY_MODE, newVal, vscode.ConfigurationTarget.Global);
         } catch (e) {
             console.error(e);
         }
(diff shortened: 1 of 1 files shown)


haneefdm:
Rejected. Sorry. All our settings will work the same way.. If you want your team to behave a certain way, that is on you. I don't even want to go into explaining to everyone why certain settings works one way vs others.

haneefdm:
Oh, I see what you did. It is still a problem and have to think of a better solution

malsyned:
> All our settings will work the same way

This isn't quite true. This setting works differently from every other setting, because this setting is modified by code within the cortex-debug extension.

I agree, there's not really any solution to this that's both good and simple. In my opinion, if you have to hard-code the location where the setting is changed, user settings are better than workspace settings. But not hard-coding it is better still.

It's possible to use the `inspect()` method to find out if there's already a setting for this somewhere and to change it in that location if so. If I send in a PR that works that way, and defaults to modifying the user's global settings if it isn't set anywhere, would you be interested in merging that?

haneefdm:
> It's possible to use the inspect() method to find out if there's already a setting for this somewhere and to change it in that location if so. If I send in a PR that works that way, and defaults to modifying the user's global settings if it isn't set anywhere, would you be interested in merging that?

Yes, I would be interested in that.. That would be much better Besides, if the user had it at WS or Folder and we are only modifying the Global setting, it would not have worked anyways.

* If the user had a setting for a specific scope, modify that one
* If not, set it at Global scope.

It is unfortunate this setting works this way. Also, when this code was written, the update/inspect did not have all these options. Even the concept of a WS did not exist or was brand new and had many issues.

malsyned:
I'm closing this pull request, it's superseded by #1109 which implements the more complicated behavior we discussed here.

Issue #717, opened 2022-08-15
Add support for envFile/dotEnv/.env files


seppestas:
Many launch configurations providers, e.g [cppdbg](https://code.visualstudio.com/docs/cpp/launch-json-reference#_environment-variable-definitions-file), [nodejs](https://code.visualstudio.com/docs/nodejs/nodejs-debugging#_load-environment-variables-from-external-file) and [Python](https://code.visualstudio.com/docs/python/debugging#_envfile), have an option to provide an envFile that contains a list of environmental variables.

This feature is very handy to transfer information between build tasks and the launch configuration since vscode is currently lacking a decent built-in feature that allows this. Please consider supporting this in cortex-debug launch configurations. It would be especially handy if this feature could be used to set things like the debugger server, device, and the executable.

My use-case is a project with many build configurations and targets. Currently a very big (15k lines and growing) launch configuration is used, which is quickly getting unmaintainable.

haneefdm:
Well, how would that be used in an embedded processor? It has no concept of an `env`. The env in those other debuggers are for the program being launched./debugged

How else are the variables that go into the env file used? Are they used to expand env vars referred to in the tasks/launch.json? As far as I know, they aren't used by anyone except the program being debugged.

You can already use env variables in your launch.json (not coming from the envFile though). Please note that we do not interpret the launch.json for variable substitutions. VSCode does that and the source of the variables can be your env, workspace, input, a task, etc.

https://code.visualstudio.com/docs/editor/variables-reference
https://code.visualstudio.com/docs/cpp/launch-json-reference#_environment-variable-definitions-file

Can you show me an example of how the envFile is used to make launch.json easier in these other debuggers? What triggers the substitutions? For a normal env. variable for instance, VSCode uses the syntax like `${env:USERNAME}`

https://code.visualstudio.com/docs/editor/variables-reference#_environment-variables

seppestas:
I admit it is a bit of a hack. The problem with ${env:} variables is that they only show VSCode's environment at the time of launch, they can't be altered by e.g tasks. The only way to change it is to launch VSCode with a different environment. Really what is missing is a decent project variable concept in VSCode, see https://github.com/microsoft/vscode/issues/154802.

I thought the plugin might be able to load variables from a .env file and substitute the ${env:<variable>} or $<variable>, but looking at the source this is indeed probably not possible because VSCode handles the expansion. 

Maybe having pre-defined env vars set some of the configuration of cortex-debug would work?

[...]

PR #1117, opened 2025-04-12
Add support for environment variable substitution in launch.json

SonnyDinnetz:
Relating to this issue: https://github.com/Marus/cortex-debug/issues/717

I've added code which checks for optional properties "envFile" and "environment" in launch.json at the beginning of resolveDebugConfiguration. I tested on Windows 10 only as I don't have access to a Linux machine rn.

1. It first checks if the user has defined an envFile and if so parses it for key/value pairs. 
2. After looking at the file it then appends anything it found to the "environment" property. 
3. It then compares each value for each key in launch.json for values that contain ${env:key_name} and if it finds them and there was a corresponding key in either "environment" or the envFile it replaces ${env:key_name} with the associated value.

My primary reason for doing this was to make automating my build process easier. My build system organizes different builds under ./build/{software version}/{gcc output} with my makefile handling everything involved there. I've grown really fond of building/flashing/debugging all with one button click in vs code, but to do that I've had to keep the file paths in my launch.json in sync with the software version defined in my makefile, but that's rather inconvenient.

With this I can create a .env file in my .vscode folder and define any values that are unique to my setup and/or the current version in it, then .gitignore the .env file so that my coworkers can maintain their own unique environment. This shouldn't have any effect on any projects which don't use the "envFile" or "environment" properties as it only makes changes to the launch config if those values are present.

Formatting for the key/value pairs:
- Any file path string which resolves to a file that can be parsed.
- File cannot begin with a comment or the regex on line 308 of configprovider.ts won't find matches (line.match(/^([^#\s]+)=(.*)$/))
- Cannot use ${workspaceFolder} to locate the file due to this function being called before VS code does it's substitution (if we wait until afterwards then we lose the placeholders from the value strings in launch.json i.e. ${env:KEY})

Example launch configuration:
```
{
    "name": "Build & Debug - J-Link", 
    "envFile": "./.vscode/.env",
    "cwd": "${workspaceFolder}",
    "type": "cortex-debug",
    "executable": "${workspaceFolder}/build/${env:software_revision}/${env:software_revision}_DEBUG.elf",
    "request": "launch",
    "servertype": "jlink",
    "serverpath": "${env:JLINK_GDB_PATH}",
    "gdbPath": "${env:ST_GDB_PATH}",
    "device": "STM32G431KBTx",
    "runToEntryPoint": "main",
    "svdFile": "${env:SVD_PATH}",
    "svdPath": "${env:SVD_PATH}",
    "liveWatch": {
        "enabled": true,
        "samplesPerSecond": 4
    },
    "v1": false,
    "serverArgs": [
        "-speed", "8000",
        "-vd"
    ],
    "interface": "swd",
    "ipAddress": "192.168.1.50",
    "preLaunchTask": "Make Debug"
}
```

Example .env file (with fake key obv):
```
software_revision="0.01.05A"
ST_GDB_PATH="C:/ST/STM32CubeCLT_1.15.0/GNU-tools-for-STM32/bin/arm-none-eabi-gdb.exe"
JLINK_GDB_PATH="C:/Program Files/SEGGER/JLink_V796s/JLinkGDBServerCL.exe"
SVD_PATH="C:/ST/STM32CubeCLT_1.15.0/STMicroelectronics_CMSIS_SVD/STM32G431.svd"

AES_PRIVATE_KEY="ABCDEFGHIJKLMNOP"
```

### DIFF
--- a/src/frontend/configprovider.ts
+++ b/src/frontend/configprovider.ts
@@ -56,6 +56,8 @@ export class CortexDebugConfigurationProvider implements vscode.DebugConfigurati
         config.gdbServerConsolePort = GDBServerConsole.BackendPort;
         config.pvtAvoidPorts = CDebugSession.getAllUsedPorts();
 
+        this.loadEnvironmentVariables(config, folder);
+
         // Flatten the platform specific stuff as it is not done by VSCode at this point.
         switch (os.platform()) {
             case 'darwin':
@@ -292,6 +294,46 @@ export class CortexDebugConfigurationProvider implements vscode.DebugConfigurati
                 delete obj[prop];
             }
         }
+    }
+
+    private loadEnvironmentVariables(config: vscode.DebugConfiguration, folder: vscode.WorkspaceFolder ) {
+        // First check for an environment file
+        if (config.envFile) {
+            const envFilePath = path.isAbsolute(config.envFile) ? config.envFile : path.join(folder?.uri.fsPath || '', config.envFile); // Resolve it with the workspace folder if needed
+            if (fs.existsSync(envFilePath)) {                               // Check if it actually exists
+                const envContent = fs.readFileSync(envFilePath, 'utf8');    // Load the environment file in as a big string to make life easy
+                const envVariables: { [key: string]: string } = {};         // Create a place to put the key/value pairs while we work on the file
+
+                envContent.split(/\r?\n/).forEach(line => {                 // Split by newlines and for each element
+                    const match = line.match(/^([^#\s]+)=(.*)$/);           // Were looking for "SYMBOL=variable" ignoring comments
+                    if (match) {
+                        const key = match[1].trim();                        // Get rid of white space
+                        let value = match[2].trim();
+                        if (value.startsWith('"') && value.endsWith('"')) { // Get rid of quotation marks if they're there
+                            value = value.slice(1, -1);
+                        }
+                        envVariables[key] = value;                          // Note that key/value pair down for use later
+                    }
+                });
+
+                config.environment = { ...config.environment, ...envVariables }; // Append it all to out environment variables.
+            } else {
+                vscode.window.showWarningMessage(`The envFile "${envFilePath}" does not exist.`); // If user provided an envFile but it didn't resolve correctly
+            }
+        }
+
+        // Now lets replace anything in launch.json that needs replaced.
+        const environment = config.environment || {};   // Ensure we can iterate over config.environment even if it's empty
+        const replaceEnvVariables = (obj: any) => {
+            for (const key in obj) {                    // Check every key
+                if (typeof obj[key] === 'string') {     // If it's a string we want to look for our search pattern
+                    obj[key] = obj[key].replace(/\${env:(\w+)}/g, (match, p1) => environment[p1] || match);
+                } else if (typeof obj[key] === 'object') { // If it's an array of some sort we want to recurse into it
+                    replaceEnvVariables(obj[key]);
+                }
+            }
+        };
+        replaceEnvVariables(config);
     }
 
     private validateLoadAndSymbolFiles(config: vscode.DebugConfiguration, cwd: string) {
--- a/debug_attributes.md
+++ b/debug_attributes.md
@@ -31,6 +31,8 @@ If the type is marked as `{...}` it means that it is a complex item can have mul
 | cwd | string | Both | Directory to run commands from |
 | debuggerArgs | array | Both | Additional arguments to pass to GDB command line |
 | device | string | Both | Target Device Identifier |
+| envFile | string | Both | Path to a file which contains environment variables for substitution in launch.json. |
+| environment | object | Both | Support for Environment Variable Substitution in launch.json |
 | executable | string | Both | Path of executable for symbols and program information. See also `loadFiles`, `symbolFiles` |
 | gdbPath | string | Both | This setting can be used to override the GDB path user/workspace setting for a particular launch configuration. This should be the full pathname to the executable (or name of the executable if it is in your PATH). Note that other toolchain executables with the configured prefix must still be available. |
 | gdbTarget | string | Both | For externally (servertype = "external") controlled GDB Servers you must specify the GDB target to connect to. This can either be a "hostname:port" combination or path to a serial port |
--- a/package-lock.json
+++ b/package-lock.json
@@ -975,6 +975,123 @@
                 "typescript": ">=4.8.4 <5.8.0"
             }
         },
+        "node_modules/@typescript-eslint/eslint-plugin/node_modules/@typescript-eslint/scope-manager": {
+            "version": "8.20.0",
+            "resolved": "https://registry.npmjs.org/@typescript-eslint/scope-manager/-/scope-manager-8.20.0.tgz",
+            "integrity": "sha512-J7+VkpeGzhOt3FeG1+SzhiMj9NzGD/M6KoGn9f4dbz3YzK9hvbhVTmLj/HiTp9DazIzJ8B4XcM80LrR9Dm1rJw==",
+            "dev": true,
+            "license": "MIT",
+            "dependencies": {
+                "@typescript-eslint/types": "8.20.0",
+                "@typescript-eslint/visitor-keys": "8.20.0"
+            },
+            "engines": {
+                "node": "^18.18.0 || ^20.9.0 || >=21.1.0"
+            },
+            "funding": {
+                "type": "opencollective",
+                "url": "https://opencollective.com/typescript-eslint"
+            }
+        },
+        "node_modules/@typescript-eslint/eslint-plugin/node_modules/@typescript-eslint/types": {
+            "version": "8.20.0",
+            "resolved": "https://registry.npmjs.org/@typescript-eslint/types/-/types-8.20.0.tgz",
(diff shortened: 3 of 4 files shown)


src/frontend/configprovider.ts
```diff
@@ -294,6 +296,46 @@ export class CortexDebugConfigurationProvider implements vscode.DebugConfigurati
...
 
+    private loadEnvironmentVariables(config: vscode.DebugConfiguration, folder: vscode.WorkspaceFolder ) {
+        // First check for an environment file
+        if (config.envFile) {
+            const envFilePath = path.isAbsolute(config.envFile) ? config.envFile : path.join(folder?.uri.fsPath || '', config.envFile); // Resolve it with the workspace folder if needed
+            if (fs.existsSync(envFilePath)) {                               // Check if it actually exists
+                const envContent = fs.readFileSync(envFilePath, 'utf8');    // Load the environment file in as a big string to make life easy
+                const envVariables: { [key: string]: string } = {};         // Create a place to put the key/value pairs while we work on the file
+
+                envContent.split(/\r?\n/).forEach(line => {                 // Split by newlines and for each element
```
haneefdm, line 307: No forEach's. Use `for ... of ...` instead. While you will see those in our old code, we don't use forEach anymore. Many reasons, most important of all single-stepping convenience.

haneefdm:
I am a bit confused about the envFile and what other debuggers do with it. It is generally meant for environment variables for the debugger or the target and not for replacements of other stuff in launch.json. The launch.json is just a VSCode thing. I the package.json, the documentation is inadequate for what envFile/environmet does and the precedence rules. See

https://code.visualstudio.com/docs/python/environments#_environment-variables
https://code.visualstudio.com/docs/cpp/launch-json-reference#_environment

Using the same names (envFile/environment) for a totally different purpose can raise support issues and is confusing. Even to me because I cross boundaries between embedded and non-embedded environments.

I haven't done a deep dive on what other do but reading from their docs the meaning does not seem to be the same.

How does work with `preLaunchTask` if it modifies the envFile? See https://github.com/microsoft/vscode-cpptools/issues/5060#issuecomment-747339576

The whole reason people were asking for this is so that their build can populate the envFile. The way this is done, how can it work? Did you test it? What does cpptools do?

haneefdm:
Your lint is failing. Please run `npm run lint` to make sure. Also do a `npn run package` so there are no build issues and that it packages properly

src/frontend/configprovider.ts
```diff
@@ -294,6 +296,46 @@ export class CortexDebugConfigurationProvider implements vscode.DebugConfigurati
...
+
+                config.environment = { ...config.environment, ...envVariables }; // Append it all to out environment variables.
+            } else {
+                vscode.window.showWarningMessage(`The envFile "${envFilePath}" does not exist.`); // If user provided an envFile but it didn't resolve correctly
+            }
+        }
+
+        // Now lets replace anything in launch.json that needs replaced.
+        const environment = config.environment || {};   // Ensure we can iterate over config.environment even if it's empty
+        const replaceEnvVariables = (obj: any) => {
```
haneefdm, line 327: This algorithm is O(n*m) isnt int? We can do this in O(n+m).  Imagine a 100 env. vars and a 100 properties in the launch config.

haneefdm:
So, who handles the actual environment variables? Maybe I missed it but are you merging the process.env with what we find in the launch config? I would start with (a copy of) process.env as the initial environment.

I am assuming VSCode uses process.env as well. And have clear/documented precedence rules.

haneefdm:
What do you know of case sensitivity/preservation nature of env vars on Windows VSCode? What Windows does may not matter. What matters is what does VSCode do on Windows. We have to emulate whatever that behavior is.

src/frontend/configprovider.ts
SonnyDinnetz: changed

src/frontend/configprovider.ts
SonnyDinnetz: I've been thinking about how to make this more efficient and wasn't sure how to get to O(n+m). Do you have an idea you can point me at?

My new strategy is to check each key in launch.json for ${env:*} noting them down as I go. Then in a separate loop check if they're in the envFile and replace them then. Should reduce the number of comparisons I'm doing a lot but I don't know if it actually changes the bigO.

SonnyDinnetz:
> What do you know of case sensitivity/preservation nature of env vars on Windows VSCode? What Windows does may not matter. What matters is what does VSCode do on Windows. We have to emulate whatever that behavior is.

Wasn't able to find documentation that outright said whether or not VSCode was intended to obey case sensitivity for env variables but I did check and at least for the ${workspaceFolder} substitution it did so I went ahead and made sure that my search considered case before replacing.

SonnyDinnetz:
> I am a bit confused about the envFile and what other debuggers do with it. It is generally meant for environment variables for the debugger or the target and not for replacements of other stuff in launch.json. The launch.json is just a VSCode thing. I the package.json, the documentation is inadequate for what envFile/environmet does and the precedence rules. See

I'll get more and clearer docs made up. I'll also get something prepared for the wiki too so I can go into details without cluttering the package.json too much.

> https://code.visualstudio.com/docs/python/environments#_environment-variables https://code.visualstudio.com/docs/cpp/launch-json-reference#_environment
> 
> Using the same names (envFile/environment) for a totally different purpose can raise support issues and is confusing. Even to me because I cross boundaries between embedded and non-embedded environments.
> 
> I haven't done a deep dive on what other do but reading from their docs the meaning does not seem to be the same.

The documentation for other implementations of these are super lacking and it seems like they all handle it differently. For example:

https://code.visualstudio.com/docs/python/environments#_environment-variables
This implementation seems to treat envFiles almost like lumps of python constants to swap out. In the example they have I think of swapping dev.env for prod.env in the launch config the same as changing the import itself.

https://code.visualstudio.com/docs/cpp/launch-json-reference#_environment
Then in this one they give next to no explanation at all but the syntax they use is completely different. Defining the key and value entirely separately using the keys "name" and "value" to define each key and value.

https://code.visualstudio.com/docs/reference/variables-reference
This was what I was using as my inspiration for environment variables. They of course don't come out and say it but they treat USERNAME here as if they were drawing it from the system variables ala powershell>echo ${env:USERNAME}. My thought process is that I don't want to define a ton of system variables on my machine when I really only want them to exist project to project so I create a file which can be dedicated to a single project or even launch config.


> How does work with `preLaunchTask` if it modifies the envFile? See [microsoft/vscode-cpptools#5060 (comment)](https://github.com/microsoft/vscode-cpptools/issues/5060#issuecomment-747339576)

I played around with placing the whole routine in resolveDebugConfigurationWithSubstitutedVariables instead of resolveDebugConfiguration but that runs into supercession issues with the actual system variables. Essentially, some time between resolveDebugConfiguration and resolveDebugConfigurationWithSubstitutedVariables VSCode is doing it's own substitution which would replace any instance of ${env:\*} that it doesn't find with an empty string making the searches and swaps done in my function always fail. What this guy https://github.com/microsoft/vscode-cpptools/issues/5060#issuecomment-747339576 is saying fits what I'm seeing. In my implementation preLaunch tasks should have no effect on the envFile or environment since by the time they run they should have already been parsed and swapped out. 
I'm definitely open to trying to find a solution for this though, maybe a different key than "${env:*}" that way windows doesn't touch it before we do? 

> The whole reason people were asking for this is so that their build can populate the envFile. The way this is done, how can it work?

My use case is entirely in swapping out aspects of the launch configurations. Moreso using the envFile to populate aspects of my builds than the the build populating the envFile.

> Did you test it?

I tested a bunch on windows, I'm planning to do some runs on a linux box this weekend to make sure it acts like I expect it to. 

> What does cpptools do?

As far as I can tell it doesn't interact with cpptools at all. What drove me to do this in the first place is that when I have cortex-debug selected as the type for a launch config it won't allow me to use the "environment" or "envFile" keywords at all in the launch configs. I think it warrants some digging though so I'll find some time to build a blinky project that uses the cpptools "environment" keyword and see if maybe I'm missing something and if it integrates well with the changes I'm asking for.

haneefdm:
> What does cpptools do?

>> As far as I can tell it doesn't interact with cpptools at all

That is not what I meant. Of course cpptools will not be involved if the debug type is `cortex-debug`. What I meant is that they have a similar feature -- how to they process it. It could be good to use their model if applicable. I would not create an embedded program to test. I would use a normal desktop C program.

I know it is easy to simply do what is needed for "my" use case. But, we have to look at the general picture. We are kind of doing this backwards. We are deriving specifications/requirements while developing. Even when I want to do something, I submit a proposal (using an issue here) and solicit feedback before writing any code.

> Wasn't able to find documentation that outright said whether or not VSCode was intended to obey case sensitivity for env variables but I did check and at least for the ${workspaceFolder} substitution it did so I went ahead and made sure that my search considered case before replacing.

Welcome to my world. Most of the time, I have to reverse engineer stuff like this. I would not use `${workspaceFolder}` to test something like this. The rules are specific to `${env:...}`

I don't want to end up with two sets of rules and two processors. We need to deal with ALL env vars. No half and half.

And, thanks for taking the initiative. I knew it would not be as simple and if I could achieve the true goal of WHY people were asking for it.

SonnyDinnetz:
I'm gonna close this because I'm just not finding the time to come back and do the work I promised. I'll try to follow up in the future, when I do that do you want me to open this PR back up or would you rather me raise an issue first?

PR #1129, opened 2025-05-27
Increase GDB Server init timeout

J-Rios:
Increasing timeout value for GDB Server connection from 10s to 30s.


haneefdm:
Justification?

Issue #978, opened 2024-02-07
Support explicit hardware breakpoints (send "hbreak" instead of "break" to GDB)


recursivenomad:
My hardware has 4 hardware breakpoints, so I pass `set remote hardware-breakpoint-limit 4` to GDB during launch via my launch.json.  However, if I set more than 4 breakpoints, GDB does not guard me from setting too many breakpoints as expected; instead it reports an error from the remote target that it "Cannot insert hardware breakpoint 5", and that "A serious error occurred with gdb, unable to continue or interrupt".

If I instead explicitly set 4 hardware breakpoints by manually passing `hbreak` to GDB, then attempting to set a 5th hardware breakpoint this way, GDB prevents me from setting this breakpoint and gracefully reports "Hardware breakpoints used exceeds limit", with no fatal errors.

When setting breakpoints in the UI, I can confirm that once they are sent from GDB to the gdbserver; seen with `set debug remote 1`, GDB is appropriately passing `Z1` commands to automatically assign hardware breakpoints.  But oddly enough, these automatic hardware breakpoints don't appear to count towards the hardware-breakpoint-limit.  Upon querying `info break`, all breakpoints are listed as `breakpoint`, not `hw breakpoint`.  This is contrary to what `info break` reports when manually passing `hbreak` to GDB.

The ideal solution for my use-case would be to allow users to specify how Cortex-Debug communicates breakpoints to GDB (specifically so I can set it to send `hbreak` instead of `break`).

This is already [a feature in Microsoft's cpptools extension](https://github.com/microsoft/vscode-cpptools/issues/4745#issuecomment-990174224), however I would prefer not to rely on cpptools as it is largely closed-source, and I am attempting to maintain a fully FOSS development environment.  I also do not know if that feature of the cpptools extension is even compatible with Cortex-Debug.

The following alternative solutions are not ideal workarounds for my use-case:
- Microsoft's cpptools extension (closed-source)
- Manually submitting `hbreak` to GDB (cumbersome, doesn't update UI)
- Intentionally avoiding setting more than 4 breakpoints (cumbersome, risks errors, unintuitive to new users)

For context, I am using Cortex-Debug with arm-none-eabi-gdb through OpenOCD to a Raspberry Pi Pico, communicating through a Picoprobe (an additional Pico running as a CMSIS-DAP debugger).

Thank you for developing and maintaining these extensions :)

litui:
[...]

Issue #555, opened 2021-12-27
How to insert hardware breakpoints？


hongshui3000:
I use the following configuration，But the gui can still only set software breakpoints。Unless I use the gdb command hbreak,
launch.json
```
            "hardwareBreakpoints": {
                "require": true,
                "limit": 2
             }
```

PhilippHaefele:
Hi,
can you write a little bit more about your setup please. Complete launch configuration would be nice.

As the consideration what kind of breakpoints is used when issuing a GDB `-break-insert` is upto the GDBServer (when not using the `-h` option), we really need that information to help you.

haneefdm:
Almost always what you get is a HW breakpoint when your program is in FLASH. gdb doesn't even know how else to set a breakpoint. GDB-servers also do not mess with FLASH on a byte basis. GDB knows with help with the gdb-server the memory types your devices have and does the right thing.

If your program is running from RAM, then that is different. So, what is the question here?

Also, the JSON snippet you posted in your `launch.json`? That is probably for the Microsoft debugger. Not ours. But you are welcome to use their debugger.

hongshui3000:
launch.json：
```
    {
            "name": "Openocd Attach",
            "type": "cortex-debug",
            "request": "attach",
            "servertype": "openocd",
            "cwd": "${workspaceRoot}",
            "executable": "./build/zephyr/zephyr.elf",
            "configFiles":[
                "board/xtensa-x-jlink.cfg",
                "board/xtensa-core-x.cfg"
            ],
            "debuggerArgs":[
              "--xtensa-core=hifi4vfp7",
              "-x", "${workspaceFolder}/ZCAN-sdk/scripts/gdb.init"
            ],
            "gdbPath":"C:\\usr\\xtensa\\XtDevTools\\install\\tools\\RI-2021.6-win32\\XtensaTools\\bin\\xt-gdb.exe",
            //"preLaunchTask": "build",
            "runToEntryPoint": "main",
            "hardwareBreakpoints": {
                "require": true,
                "limit": 2
             }
        }
```
Where should "hardwareBreakpoints" be placed? my program is in FLASH

haneefdm:
Oh, is this because you are using a non-ARM (Cortex) device?

[...]

PR #1166, opened 2025-11-07
Feature: forced hardware breakpoint support

litui:
This feature adds the configuration option `gdbBreakpointType` with a default value of `auto`.

When set to `hardware` it will inject the `-h` option on all gdb-server `break-insert` calls as per guidance found [here](https://sourceware.org/gdb/current/onlinedocs/gdb.html/GDB_002fMI-Breakpoint-Commands.html).

There is an argument to be made that this support is a useful addition to facilitate stopgap cortex-debug use with incomplete gdb-server implementations or incorrect memory maps.

The argument has been made (and rejected by the maintainers) on the following issues, that I have encountered:
- #555 
- #978 

Submitting as a draft PR not necessarily for acceptance but to provide an easy-to-find option for others who may need this functionality.

I have not yet tested these changes beyond confirmation that the extension builds and the vsix can be installed into VSCode.

### DIFF
--- a/src/backend/mi2/mi2.ts
+++ b/src/backend/mi2/mi2.ts
@@ -633,7 +633,12 @@ export class MI2 extends EventEmitter implements IBackend {
         }
         return new Promise((resolve, reject) => {
             let bkptArgs = '';
-            bkptArgs += this.extraBreakpointArgs;
+
+            // Only insert extraBreakpointArgs if there is no logMessage as dprintf-insert
+            // does not support the -h argument.
+            if (!breakpoint.logMessage) {
+                bkptArgs += this.extraBreakpointArgs;
+            }
 
             if (breakpoint.hitCondition) {
                 if (breakpoint.hitCondition[0] === '>') {


recursivenomad:
I just built and tested this - it works perfectly!! 🙏

After setting `"gdbBreakpointType": "hardware"` in my `launch.json`, now when I try to add a breakpoint which is greater than `hardware-breakpoint-limit`, it super gracefully and immediately shows a grey circle, no exceptions are thrown, and when I hover over the grey circle it very clearly says "Hardware breakpoints used exceeds limit".  Sending `info break` to GDB now also correctly lists each breakpoint as a "hw breakpoint".  This is exactly the behaviour I have been seeking, as described in #978!

| . | . |
|---|---|
| image | image |

Thank you so much for taking the time to compose this fix!

recursivenomad:
No worries, I'm in the same boat.  In hindsight, I probably spent far more time familiarizing with GDB's codebase than it would have taken to learn VSCode extension development for this - but I'm very grateful that you took the plunge 🙂

haneefdm:
You only half modified the package.json? Did not address the `attach` side of it.

But, I just commited changes that forces HW breakpoints as well as enforce a limit. One without the other is only half a solution. I will close this PR without merging soon.

PR #1174, opened 2025-12-08
Add support for custom gdbHost in jlink servertype

sigurdal:
This adds support for specifying the host for GDB when running with "servertype": "jlink".

This is to allow running the build environment and GDB runs in WSL, while having JLinkGDBServerCL run in Windows and thus avoiding mapping the USB device into WSL.

If not set, it defaults to localhost to maintain backwards compatibility.

In my setup I have the following lines in settings.json:
```json
{
    "cortex-debug.JLinkGDBServerPath": "/mnt/c/Program Files/SEGGER/JLink/JLinkGDBServerCL.exe",
    "cortex-debug.gdbPath": "/path/to/bin/arm-none-eabi-gdb"
}
```

And add the following lines to launch.json:
```json
"gdbHost": "172.X.Y.1",
"serverArgs": [
    "-nolocalhostonly",
],
```

But as the WSL IP might change, it can be automatically updated, for example by using the `augustocdias.tasks-shell-input` extension and:
```json
"configurations": [
    {
        "name": "Debug with J-Link",
        "gdbHost": "${input:wslHostIp}",
        "serverArgs": [
            "-nolocalhostonly",
        ],
        ...
    }
],
"inputs": [
    {
        "id": "wslHostIp",
        "type": "command",
        "command": "shellCommand.execute",
        "args": {
            "command": "ip route show | grep -i default | awk '{ print $3}'",
            "cwd": "${workspaceFolder}",
            "useSingleResult": true
        },
    }
]
```

### DIFF
--- a/debug_attributes.md
+++ b/debug_attributes.md
@@ -32,6 +32,7 @@ If the type is marked as `{...}` it means that it is a complex item can have mul
 | debuggerArgs | array | Both | Additional arguments to pass to GDB command line |
 | device | string | Both | Target Device Identifier |
 | executable | string | Both | Path of executable for symbols and program information. See also `loadFiles`, `symbolFiles` |
+| gdbHost | string | Both | Host for the GDB server connection (defaults to localhost). Only used for servertype jlink. |
 | gdbInterruptMode | string | Both | Whether GDB shall be interrupted using "exec-interrupt" (default) or by signaling "SIGINT" |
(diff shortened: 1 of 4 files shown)


sigurdal:
This PR only enabled the option of using something else than localhost as the target when the servertype was set to jlink, with WSL as an example usage scenario where it could be useful.

As the PR cannot be accepted I have closed it.

haneefdm:
Full wsl/docker/ssh support is coming, this part is already implemented/working. But in a new repo...

https://github.com/mcu-debug/mcu-debug/blob/756c2e7ca868f1e006183f62bbeddd43179884a7/TaskList.md?plain=1#L68

Not ready for even a beta release, complete rewrite.

PR #1217, opened 2026-04-23
Allow loading with -device file=mybin.elf

simonsso:
For some reason my qemu wants to be loaded with `-device loader,file=executable.elf` instead of `-kernel executable.elf` this option solved my need.

### DIFF
--- a/src/common.ts
+++ b/src/common.ts
@@ -275,6 +275,7 @@ export interface ConfigurationArguments extends DebugProtocol.LaunchRequestArgum
     gdbInterruptMode: GDBInterruptMode;
     objdumpPath: string;
     serverArgs: string[];
+    serverOptionLoadWithDeviceFile: boolean;
     serverCwd: string;
     device: string;
     loadFiles: string[];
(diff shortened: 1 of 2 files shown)


haneefdm:
Can't you use `serverArgs` in launch.json? Or use "external" server type as an option to start the server externally.

Our package.json has grown to be too large to be maintainable. If there are already ways for you to use existing mechanims, we should do so.

What have you tried?

simonsso:
I could have added it to serverArgs but I needed to remove `-kernel file.bin` and remove it with the `-device loader,file=file.bin` (Note empathize where spaces are and are not needed)

If the addon only adds -kernel executable if executable is set I would have been able to use only server args.

simonsso:
@haneefdm  it looks like @Marus is only on github once or twice a year and is not maintaining this actively right now. 

I will have to start maintaining my own fork, welcome over to my fork https://github.com/simonsso/cortex-debug with more comments and suggestions

simonsso:
@haneefdm I tried to make executable argument optional not to load it with -kernel, but that did not work as the symbols were not loaded for gdb.

haneefdm:
You can use `symbolFiles` to load symbols into gdb. Default is the executable but using `symbolFiles` allows you to load zero or more elf files into gdb. Some people have multiple elf files.

However, if the qemu args still uses the exectuable (despite having symbolFiles), then maybe we can do something about that.

haneefdm:
There are other ways already to do this

PR #1218, opened 2026-04-23
Serial ports redirected to vscode Output tab

simonsso:
This patch creates a named pipe for each serial port needed. Appends the same numer of -serial args to qemu and the then redirects that output to a tab in Output for vscode.

Parts of this commit is written using AI tools.


simonsso:
I think this is a great feature, but I am not a typescript developer so the code could probably be improved and rewritten. My scope was only qemu but maybe there are more usecases in this extension which could use the same mechanisms.

simonsso:
Whops my branch was overwritten. This could be fetched from my fork if wanted.

haneefdm:
Have you thought of writing this as a separate extension? You can have a debug session tracker that reacts to a debug session and can independently do its own thing. This is how I did memory-view, rtos-views and peripheral-views. They are not part of this extension. Take a look at those repos.

simonsso:
@haneefdm I did not try to write another extension, I just wanted something that worked for me, and if someone liked it I could share it.

I notice there are no releases since 2023 but code is still being updated in this repo.

Issue #1226, opened 2026-05-15
'Jump to Cursor' feature not working


basit-s:
**Describe the bug**
Current release 1.12.1 advertises support for Jump to Cursor, but it appears to not be working.

**To Reproduce**
1. Start debug session
2. Right click line of code
3. Select "Jump to Cursor"
4. Error pop-up indicates "Set next statement is not supported by the current debugger"

**Expected behavior**

The request sets the location where the debuggee will continue to run. This makes it possible to skip the execution of code or to execute code again. The code between the current location and the goto target is not executed but skipped.

**Screenshots**

<img width="659" height="185" alt="Image" src="https://github.com/user-attachments/assets/5842ae57-5d99-481f-82fb-eede648fc14f" />

**Environment (please complete the following information):**

 - Cortex-Debug Version 1.12.1
 - OS: Linux Ubuntu 24.04.4 LTS
 - GDB Version: 13.3.1
 - Compiler Toolchain Version: arm-none-eab V 13.3.Rel1

**Additional context**
I found the original feature implementation for 'Jump to Cursor' in PR #417 and Issue #418, but it seems to have been broken during a refactor in commit 1ff878c.
* 6ac5c31, “Jump to cursor, PR 417”, added Jump to Cursor support
* 1ff878c, “hardwareBreakpoints to force HW only bpts”, removed the low-level MI goto helper and changed gotoTargetsRequest to create a temporary breakpoint instead

basit-s:
If someone else could verify if it's failing on their machine with the latest cortex-debug release version, that would be appreciated.

I have raised a PR https://github.com/Marus/cortex-debug/pull/1227 which seems to resolve this issue. It seems like it was originally introduced around 2021 (see PR https://github.com/Marus/cortex-debug/pull/417 and Issue https://github.com/Marus/cortex-debug/issues/418) but may have broke in 2025 (1ff878c)

basit-s:
Marking this as closed - it seems like there was some issue with my setup as I can't reproduce the issue anymore. I tried it on my primary machine which had the issue as well as two other machines.

Issue #418, opened 2021-03-23
Support for 'Jump to Cursor' command


FabianoGK:
Visual Studio Code has a debug command _Jump to Cursor_, which lets you move program execution to a new location without executing any of the code in between. If _Jump to Cursor_ is supported by a debugger, the new command appears in the editor context menu and Command Palette while debugging.

This feature was added to Jan Jurzitza's (WebFreak) code-debug extension by WebFreak001/code-debug#177.

I created a PR #417 with the backported changes.

zrno:
Can this be closed now? It is mentioned as implemented in the latest release notes for 4.0.0 ?

haneefdm:
Oh yes. Thanks. Yes, it is in 0.4.0

PR #417, opened 2021-03-23
Add support for 'Jump to Cursor' command.

FabianoGK:
I backported this feature from [WebFreak001/code-debug](https://github.com/WebFreak001/code-debug): [src/mibase.ts](https://github.com/WebFreak001/code-debug/tree/master/src/mibase.ts#L651-L665) and [src/backend/mi2/mi2.ts](https://github.com/WebFreak001/code-debug/tree/master/src/backend/mi2/mi2.ts#L484-L495).

`supportsGotoTargetsRequest` activates the _Jump to Cursor_ command in the Command Palette of Visual Studio Code and in the source code file context menu.

PR #1227, opened 2026-05-15
Re-added support for 'Jump to Cursor' command

basit-s:
This PR restores Jump to Cursor behavior and aligns the adapter with expected DAP request flow, see Issue #1226 

I found the original feature implementation for 'Jump to Cursor' in PR #417 and Issue #418, but it seems to have been broken during a refactor in commit 1ff878c.
* 6ac5c31, “Jump to cursor, PR 417”, added Jump to Cursor support
* 1ff878c, “hardwareBreakpoints to force HW only bpts”, removed the low-level MI goto helper and changed gotoTargetsRequest to create a temporary breakpoint instead

Current master still advertises supportsGotoTargetsRequest, and the changelog still says Jump to Cursor is supported, but I cannot find a gotoRequest implementation in the adapter anymore, and the MI backend no longer has the exec-jump path that PR #417 added. So this looks like a partial DAP implementation now: gotoTargetsRequest exists, but the actual goto step appears to be missing. When running the code, Jump to Cursor returns the error message shown in the image below.

### DIFF
--- a/src/backend/mi2/mi2.ts
+++ b/src/backend/mi2/mi2.ts
@@ -565,6 +565,18 @@ export class MI2 extends EventEmitter implements IBackend {
         });
     }
 
+    public goto(filename: string, line: number): Thenable<boolean> {
+        if (trace) {
+            this.log('stderr', 'goto');
+        }
+        return new Promise((resolve, reject) => {
+            const target = `"${filename ? `${escape(filename)}:` : ''}${line.toString()}"`;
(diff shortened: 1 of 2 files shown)


basit-s:
Marking this as closed - it seems like there was some issue with my setup as I can't reproduce the issue anymore. I tried it on my primary machine which had the issue as well as two other machines.

haneefdm:
Thank you for the follow-up. But if you find it again, we can reopen

PR #1232, opened 2026-06-03
Make GDB startup timeout configurable

donnel666:
Summary:
- Add gdbStartupTimeout launch/attach configuration with schema and docs.
- Keep the default startup wait at 5000 ms while allowing slower systems to override it.
- Apply the configured timeout when waiting for the initial GDB response.
- Update the startup error message and add sanitizer coverage.

Tests:
- npm run test-compile
- npm run lint
- npx mocha --ui tdd out/test/suite/common.test.js

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # ChangeLog
 
 # V1.13.0-pre9
+* Added `gdbStartupTimeout` launch configuration to control how long Cortex-Debug waits for GDB to respond during startup.
 * Backed out change where we try "monitor exit". See [Issue #1185](https://github.com/Marus/cortex-debug/issues/1185)
 * Increased server temout to 10 mins
 * Fix for https://github.com/Marus/cortex-debug/issues/1167 Race comdition in parallel reading of multiple symbol files
--- a/debug_attributes.md
+++ b/debug_attributes.md
@@ -34,6 +34,7 @@ If the type is marked as `{...}` it means that it is a complex item can have mul
 | executable | string | Both | Path of executable for symbols and program information. See also `loadFiles`, `symbolFiles` |
 | gdbInterruptMode | string | Both | Whether GDB shall be interrupted using "exec-interrupt" (default) or by signaling "SIGINT" |
 | gdbPath | string | Both | This setting can be used to override the GDB path user/workspace setting for a particular launch configuration. This should be the full pathname to the executable (or name of the executable if it is in your PATH). Note that other toolchain executables with the configured prefix must still be available. |
+| gdbStartupTimeout | number | Both | Maximum time in milliseconds to wait for GDB to respond during startup before treating launch as failed. |
 | gdbTarget | string | Both | For externally (servertype = "external") controlled GDB Servers you must specify the GDB target to connect to. This can either be a "hostname:port" combination or path to a serial port |
 | graphConfig | {object} | Both | Description of how graphing can be done. See our Wiki for details |
 | hardwareBreakpoints | object | Both | WARNING: Force only HW breakpoints to be used. By default GDB will use HW or SW  breakpoints depending on memory type. Use this only in rare circumstances to work around issues in your gdb-server or hardware. This setting is NOT recommended for general use. |
--- a/package.json
+++ b/package.json
@@ -554,7 +554,7 @@
                                 "items": "string"
                             },
                             "gdbStartupTimeout": {
-                                "default": 10000,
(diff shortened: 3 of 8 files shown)


haneefdm:
What problem are you solving? Why does gdb even need a timeout just to start. It isn't doing any work. That is like having a timeout for `ls`

We originally added a timeout for bad linux installations and once installed properly it was never needed. We could not tell the difference between gdb hanging infinitely vs it doing sometthing and we had to notify the user that something is wrong. Not because somehow magically gdb every came back as started running. Also, for those whose gdb is working properly, the 5 or 10 second timeout is meaningless. It happens instantly. We don't even load symbols

So, I want to understand what is going one. I don't take adding to our huge package.json lightly.

donnel666:
> What problem are you solving? Why does gdb even need a timeout just to start. It isn't doing any work. That is like having a timeout for `ls`
> 
> We originally added a timeout for bad linux installations and once installed properly it was never needed. We could not tell the difference between gdb hanging infinitely vs it doing sometthing and we had to notify the user that something is wrong. Not because somehow magically gdb every came back as started running. Also, for those whose gdb is working properly, the 5 or 10 second timeout is meaningless. It happens instantly. We don't even load symbols
> 
> So, I want to understand what is going one. I don't take adding to our huge package.json lightly.

As an alternative, if you prefer not to add another configuration option, I can change the PR to only increase the fixed startup timeout to 20 or 30 seconds. For working GDB installations there should be no
  practical impact, since the timer is cleared immediately once GDB responds. The only downside is that truly broken startup cases would report the error later, but waiting 20 or 30 seconds for a startup
  failure seems acceptable and would avoid adding another package.json setting.

haneefdm:
No, I can do it. But anytime we have a large timeout, we have to periodically let people know we are still waiting and we ourselves are not hung.

In your case though don't you think 10 seconds was long enough for someone to realize gdb is not working. What would 30 seconds do?

donnel666:
In my case, 10 seconds was not long enough. The EIDE bundled GDB on my Windows setup can take more than 10 seconds before the first MI response. With the current 5 second timeout it fails, and with a 10
  second timeout it can still fail. When I changed the local timeout to 20 seconds, the same GDB executable started successfully and the debug session continued normally.

  So 30 seconds is not meant to prove that GDB is broken. It is meant to avoid classifying a slow-but-working GDB startup as broken. For normal GDB installations it should not change anything, because the
  timer is cleared as soon as GDB responds. It only affects the slow startup or failure path.

  I agree with your point about feedback. If the timeout is increased to 20 or 30 seconds, printing a message every 5 seconds like “Still waiting for GDB to start...” would make sense, so users know Cortex-
  Debug itself is not hung.

  For my case, 20 seconds is enough based on testing. I mentioned 30 seconds only as a safer upper bound for similar Windows setups where startup time may vary due to toolchain startup, DLL loading, Python
  initialization, or antivirus scanning. I am fine with whatever value you think is appropriate, as long as it is above the 10+ seconds needed by this working setup.

haneefdm:
Yes, I will do 30 secs but with a message every 5 seconds. This is what we do with the gdb-server as well. But those have real issues. Your suffering with such a gdb is just a shame. Yes, python, antivirus but even then 20 seconds is wild. Good luck. I hope it gets better for you. You ever try WSL? Even you compiles can go 5-10x faster. But issues I am sure with USB, but some people have had luck with USBIP.

I am working on a native solution for WSL but it will be a little while before being ready.

donnel666:
Thanks, that sounds good to me. A 30 second timeout with a message every 5 seconds would solve my case, and the periodic message should make it clear that Cortex-Debug itself is still alive.

  I do use WSL for many development tasks, and I agree that builds can be much faster there. The hard part for me is embedded debugging. In my setup, moving the whole workflow to WSL has been difficult because
  of USB debug probes, J-Link driver support, USB/IP setup, and vendor tooling. Installing and reliably using J-Link inside WSL is already a non-trivial part of the migration.

  I did try moving this workflow to WSL before, but I could not get a reliable setup for embedded debugging. If the WSL ecosystem and probe support become smoother, that would definitely remove a lot of pain.
  For now, though, staying on native Windows is still the lower-friction path for this particular workflow.

  I would be happy to test the 30 second startup timeout change. I am also interested in the native WSL solution you mentioned when it becomes ready.

haneefdm:
Watch for it here...

https://mcu-debug.github.io/mcu-debug/docs/remote

haneefdm:
I am closing this

